PR #7, opened 2019-04-01
Shipping split fixes, docs

pawelpolewicz:


ppolewicz:
- `contrib/bash_completion` removed
- the structure of the API should change in v1.0.0, we wanted to get everything working in the current structure and then design a new interface once we have the testing harness and everything associated with it
- the check for copyrights will be improved in the next PR (I already wrote it, but do not want to pollute this PR so that it can be reviewed properly)
- `DownloadTests` in `test_bucket.py` should not inherit from test base class - if it did, those tests would be executed. `DownloadTests` collects the test cases and later on other classes inherit from it to run the scenarios on "their" transferer class. I checked the test logs and it looks that this works correctly (we run tests for default transferer, parallel transferer and simple transferer but not for `DownloadTests` itself - as intended)

As for the docs, the listed issues as well as fixing ToC (if I can figure it out) will be done in next PRs to unblock the migration.

PR #13, opened 2019-04-06
Fix transferer crashing on empty file download attempt

ppolewicz:
Issue reported in https://github.com/Backblaze/B2_Command_Line_Tool/pull/565 with a patch provided by @oligriffiths, I just added tests and migrated the patch to the correct repository.


ppolewicz:
@bwbeach can you understand what coveralls is trying to say? Which line is no longer covered by tests?

PR #22, opened 2019-04-06
Fix ls retries

ppolewicz:
Fixes https://github.com/Backblaze/B2_Command_Line_Tool/issues/562

still needs more tests, but before I'll do those, I'd like to have some feedback. @bwbeach is this the right approach, to move the retry count setting so far up? I tried to put it in raw_api, but it doesn't seem right, library user should not need to touch it and retries seems like something that might need to be configurable in the program that uses b2sdk. B2Session is not a right place for it either, so I think it's either transferer or B2Api, but transferer is only for stream transfers and not for metadata operations, besides B2Api is usually constructing the B2Http... It seems like the least bad place to put it in.

In the future I would pass control over the retries even further, so that the creator of B2Api can provide an object which will be used to decide on case-by-case basis whether a retry should be attempted or not. This will be then respected for both streaming (in trasferer) and metadata operations, so that if the whole internet connection is down for 2 minutes, the application (such as sync or b2_fuse) might decide to just wait - if there is no internet connection then breaking a large-scale backup/restore operation will not do anyone any good (see https://github.com/Backblaze/B2_Command_Line_Tool/issues/562#issuecomment-478299210). For now, passing a number of tries seems like a good improvement.

BTW the same object could be responsible for throttling streaming operations. This would in turn allow the user to prioritize transfers, stop them (by setting throttle to zero and waiting for a timeout) and all kinds of fancy file-transfer gui application features.

### DIFF
--- a/b2sdk/b2http.py
+++ b/b2sdk/b2http.py
@@ -262,7 +262,7 @@ class B2Http(object):
         """
         self.callbacks.append(callback)
 
-    def post_content_return_json(self, url, headers, data, try_count=1, post_params=None):
+    def post_content_return_json(self, url, headers, data, try_count=5, post_params=None):
         """
         Use like this:
 
@@ -305,7 +305,7 @@ class B2Http(object):
         finally:
             response.close()
 
-    def post_json_return_json(self, url, headers, params, try_count=1):
+    def post_json_return_json(self, url, headers, params, try_count=5):
         """
         Use like this:
 
@@ -326,7 +326,7 @@ class B2Http(object):
         data = six.BytesIO(six.b(json.dumps(params)))
         return self.post_content_return_json(url, headers, data, try_count, params)
 
-    def get_content(self, url, headers, try_count=1):
+    def get_content(self, url, headers, try_count=5):
         """
         Fetches content from a URL.
 
--- a/b2sdk/b2http.py
+++ b/b2sdk/b2http.py
@@ -219,7 +219,7 @@ class B2Http(object):
     """
     A wrapper for the requests module.  Provides the operations
     needed to access B2, and handles retrying when the returned
-    status is 503 Service Unavailable or 429 Too Many Requests.
+    status is 503 Service Unavailable, 429 Too Many Requests etc
 
     The operations supported are:
        - post_json_return_json
--- a/b2sdk/b2http.py
+++ b/b2sdk/b2http.py
@@ -49,8 +49,7 @@ def _translate_errors(fcn, post_params=None):
     Calls the given function, turning any exception raised into the right
     kind of B2Error.
 
-    :param post_params: request parameters
-    :type post_params: dict
+    :param dict post_params: request parameters
     """
     try:
         response = fcn()
@@ -110,10 +109,8 @@ def _translate_and_retry(fcn, try_count, post_params=None):
     Try calling fcn try_count times, retrying only if
     the exception is a retryable B2Error.
 
-    :param try_count: a number of retries
-    :type try_count: int
-    :param post_params: request parameters
-    :type post_params: dict
+    :param int try_count: a number of retries
+    :param dict post_params: request parameters
     """
     # For all but the last try, catch the exception.
     wait_time = 1.0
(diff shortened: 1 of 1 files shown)


certainmagic:
In b2-sdk-java, we let the user provide a retry policy object. Actually they provide a factory for one-time-use policy objects, each of which can track its own retry count, etc.

ttfn,
ab

ppolewicz:
I like b2-sdk-java design more than I like the solution provided in this PR. The design has **_JAVA_** written all over it, but I think we should use it.

The user would provide an object inheriting from AbstractRetryPolicyFactory, which would register an intent in a hierarchical structure. This way we can have an object for the sync operation, another one for uploading a large file (where factory gets the sync retry policy object in _parent_ optional parameter) and another one for a part (where large file upload is a parent).

Current retry mechanism in b2-sdk-python is implemented on two layers, one retries on b2http level and another one is higher up in order to refresh upload server/token pair. @certainmagic does b2-sdk-java policy object have a choice on whether to:
1. retry the same request
2. retry but with refreshing a token
3. give up
or how is it handled there? If there is an elegant way to do it, I would try to avoid re-inventing.

certainmagic:
Hi Pawel --

Good questions.  :)

Rather than answering here, I have added to b2-sdk-java's README.md
<https://github.com/Backblaze/b2-sdk-java/blob/master/README.md>'s
"STRUCTURE" section which has info for SDK developers.  I've added info
about how we cache account authorizations and upload urls and how we do
retries with some info about how retrying affects those caches.  Please let
me know if you have further questions.

thanks for prompting me to improve the b2-sdk-java docs.  :)

ttfn,
ab

bwbeach:
I think this PR is fine for now, and I like the idea of using a policy object in the future.  Introducing the policy object will probably be an incompatible change, and will need to wait for the next major version.

ppolewicz:
If we are going to use the policy object, we should throw this PR away and make it right. I don't want us to support both the policy object and `http_try_count` on all layers, figure out which approach to use when both are provided etc, then go through deprecation process. Lets make it right.

I think the default `RetryManager` should have `register_intent()` method which would provide `SyncIntent`, then `LargeFileUploadIntent` and `FilePartUploadIntent`, latter two would have a `.parent` public attribute and other associated fields. We'd also use those intents for logging. Then, `successfull_execution()`, `transient_failure()` and `fatal_failure()` would be called, with the `RetryManager` outputting a `Retry(delay)` object or `None` if no retry is desired (so that a smart scheduler can one day avoid blocking a thread in `sleep()`, which for big syncs with many failures, retries and threads could cause a really bad experience. Does it make sense?

bwbeach:
I agree that adding a retry count, and then replacing it with a policy object adds confusion, and would be good to avoid.

What's the purpose of the `RetryManager` in your proposal?  It seems simpler to me to just provide a `RetryPolicy` object to the `B2Api`, either when it is created, or when each call is made.

The Java implementation distinguishes between failures that are "retryable" and those that are "immediately retryable", which seems like a good idea.

For now, I'm fine just changing the number from 1 to 5 while we think this through.

certainmagic:
btw, the B2RetryPolicy is also given an "operation" argument which is
"usually" the name of the b2 operation; it looks like there are less than a
handful of non-operation names in use, like 'getAccountId' and
'getDownlodByIdUrl' where we're using a cache or interface instead of
directly making an api call.

the 'operation' parameter lets the policy have operation-specific policies.

...and, again, i haven't really looked at the python sdk code in a long,
long, long time, so i'm just trying to provide info about the java design
so you have it while thinking about what's right for the python sdk.

ttfn,
ab

ppolewicz:
@certainmagic I want to implement what you mentioned but instead of a simple string that gives just a name of the operation, I'd give an object, which contains (among other things the operation name). Imagine you are running two sync operations on the same Api object and that the retry delegate decides to stop retrying a request on the first sync operation for some reason. Now that sync will fail whatever happens next, so all future requests should be terminated (even when they start)... but only those that have the ultimate parent operation (sync1) the same as the failed request. All operations that have some other source intent (sync2) should be continued. Can't do that with just the operation name.

The retry delegate can also realize there are two syncs running in parallel and it can decide to finish the first one before processing the second one, so it can start delaying the start of the tasks from sync2 group... Or it can decide that sync1 is downloading and sync2 is uploading so we should run them in parallel after all. I'm starting to think it should be called a Scheduler, and managing retries is just one of its functions.

(in the design described in my last message, `register_intent()` would need to be able to return something like "ok", delay or "abort")
I still don't have a full picture of this, but managing a sync is a problem that we've had for a long time now and retrying is only a part of it. If we solve only the retries now, it will put us in a difficult situation later when we'll be addressing the larger problem.

ppolewicz:
Ćhanged the PR to just increase the number of attempts for now, as @bwbeach suggested

PR #25, opened 2019-04-12
app key renaming

spartan218:
These are just the changes to areas where we use account ID
for authorization to be renamed.

All Tests in pre-commit.sh pass.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,13 @@ b2sdk>=0.0.0,<1.0.0
 
 # Release History
 
+## 0.1.4 (2019-04-12)
+
+Renaming accountId for authentication to application key Id
+    Note: this means account Id is still backwards compatible,
+    only the terminology has changed.
+
+
 ## 0.1.4 (2019-04-04)
 
 Initial official release of SDK as a separate package (until now it was a part of B2 CLI)
@@ -69,9 +76,9 @@ Before checking in, use the `pre-commit.sh` script to check code formatting, run
 unit tests, run integration tests etc.
 
 The integration tests need a file in your home directory called `.b2_auth`
-that contains two lines with nothing on them but your account ID and application key:
+that contains two lines with nothing on them but your application Key ID and application key:
 
-     accountId
+     applicationKeyId
      applicationKey
 
 We marked the places in the code which are significantly less intuitive than others in a special way. To find them occurrences, use `git grep '*magic*'`.
--- a/README.md
+++ b/README.md
@@ -34,8 +34,8 @@ b2sdk>=0.0.0,<1.0.0
 
 ## 0.1.4 (2019-04-12)
 
-Renaming accountId for authentication to application key Id
-    Note: this means account Id is still backwards compatible,
+Renaming account ID for authentication to application key ID
+    Note: this means account ID is still backwards compatible,
     only the terminology has changed.
 
 
@@ -76,7 +76,7 @@ Before checking in, use the `pre-commit.sh` script to check code formatting, run
 unit tests, run integration tests etc.
 
 The integration tests need a file in your home directory called `.b2_auth`
-that contains two lines with nothing on them but your application Key ID and application key:
+that contains two lines with nothing on them but your application key ID and application key:
 
      applicationKeyId
      applicationKey
--- a/b2sdk/account_info/in_memory.py
+++ b/b2sdk/account_info/in_memory.py
@@ -35,7 +35,7 @@ class InMemoryAccountInfo(UrlPoolAccountInfo):
     """
     Holder for all account-related information that needs to be kept
     between API calls, and between invocations of the command-line
-    tool.  This includes: account id, application key ID, application key,
+    tool.  This includes: account ID, application key ID, application key,
     auth tokens, API URL, download URL, and uploads URLs.
     """
 
--- a/b2sdk/account_info/in_memory.py
+++ b/b2sdk/account_info/in_memory.py
@@ -35,7 +35,7 @@ class InMemoryAccountInfo(UrlPoolAccountInfo):
     """
     Holder for all account-related information that needs to be kept
     between API calls, and between invocations of the command-line
-    tool.  This includes: account id, application key id, application key,
+    tool.  This includes: account id, application key ID, application key,
     auth tokens, API URL, download URL, and uploads URLs.
     """
 
@@ -52,7 +52,7 @@ class InMemoryAccountInfo(UrlPoolAccountInfo):
 
     def _clear_in_memory_account_fields(self):
         self._account_id = None
-        self._app_key_id = None
+        self._application_key_id = None
         self._allowed = None
         self._api_url = None
         self._application_key = None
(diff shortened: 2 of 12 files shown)


spartan218:
I think I disagree with coverage.
The area where I think it see less coverage is in 
`raw_api.py` because I changed the account ID we compare to from the response
rather than the account ID from the `b2_authorize_account` call which I believe is 
the proper way to check an account ID.

b2sdk/api.py
```diff
@@ -127,37 +127,37 @@ def authorize_automatically(self):
...
-        :type account_id_or_key_id: str
+        :param application_key_id: application key ID
+        :type application_key_id: str
         :param application_key: user's application key
         :type application_key: str
         """
         # Clean up any previous account info if it was for a different account.
         try:
-            old_account_id = self.account_info.get_account_id()
+            old_id = self.account_info.get_application_key_id()
```
spartan218: Woops this was supposed to be in a different PR

spartan218: Let me know if you feel strongly about this in a different PR and I can remove it and maybe add a test to make sure the cache is cleared?

b2sdk/account_info/sqlite_account_info.py
```diff
@@ -320,9 +320,9 @@ def get_account_id(self):
         """
         return self._get_account_info_or_raise('account_id')
 
-    def get_account_id_or_app_key_id(self):
+    def get_application_key_id(self):
         """
-        Return account ID or appliction key ID.
+        Return application key ID or Account ID.
```
ppolewicz:
```suggestion
        Return application key id.
```
How `application key id` and `account id` relate to each other, why they can be equal and when, is outside of scope of this docstring. Whoever reads this, should already know what application key id is, or should go check it in some other docs.

README.md
```diff
@@ -69,9 +76,9 @@ Before checking in, use the `pre-commit.sh` script to check code formatting, run
 unit tests, run integration tests etc.
 
 The integration tests need a file in your home directory called `.b2_auth`
-that contains two lines with nothing on them but your account ID and application key:
+that contains two lines with nothing on them but your application Key ID and application key:
```
ppolewicz: please decide on the spelling. As of now in this PR we can see it as
* application key id
* application Key ID
* application key Id

I personally prefer the first one, but the most important bit is to be consistent (so that we can one day grep the repo to find it etc)

b2sdk/account_info/in_memory.py
```diff
@@ -133,13 +133,13 @@ def get_account_id(self):
...
-    def get_account_id_or_app_key_id(self):
+    def get_application_key_id(self):
         """ 
-        Returns the account id or key id used to authenticate
+        Returns the application key id used to authenticate
 
         :rtype: str
         """
-        return self._account_id_or_app_key_id
+        return self._app_key_id
```
ppolewicz: The variable containing it is sometimes called:
* application_key_id
* _app_key_id / app_key_id

I think there is no need to shorten it (we all use variable completion in our IDEs anyway), but there is a great benefit of having it consistently named across the entire library. When the argument is renamed in each layer, it becomes difficult to read the code.

b2sdk/raw_api.py
```diff
@@ -548,9 +548,9 @@ def test_raw_api_helper(raw_api):
     this test will break and we'll have to do something about
     it.
     """
-    account_id = os.environ.get('TEST_ACCOUNT_ID')
-    if account_id is None:
-        print('TEST_ACCOUNT_ID is not set.', file=sys.stderr)
+    application_key_id = os.environ.get('TEST_APPLICATION_KEY_ID')
```
ppolewicz: this requires changes in the build environment, please make sure you ask Brian Beach to adjust the CI configuration to provide the key on both env variables during the transition period

ppolewicz:
Also, the changelog entry is missing. Please note that you've changed signature of AbstractAccountInfo (and all children) from `.get_account_id_or_app_key_id()` to `.get_application_key_id()`, and also the argumnt name was changed in `_set_auth_data`. While for the first change we could simply provide a proxy method with a deprecation warning, the second one is tougher and would require some crafty decorator (which doesn't mean it's impossible, it's just less clean).

Therefore a decision needs to be made, whether we want to bump the major version because of this change, or not.

b2sdk/account_info/sqlite_account_info.py
spartan218: Are you suggesting that I leave the original docstring or that its fine?

b2sdk/raw_api.py
spartan218: I thought that the CI configuration only supplied to strings which it mapped to environment variable names. I will check with Brian ty.

b2sdk/account_info/sqlite_account_info.py
ppolewicz: The old version was:
```
Return account ID or appliction key ID.
```
Your change is:
```
Return application key ID or Account ID.
```
and I suggest to:
```
Return application key id.
```
(if you want to accept my suggestion, you can click a button on github)

b2sdk/raw_api.py
ppolewicz: Yes. For the transition period we should export this as **both** `TEST_ACCOUNT_ID` and `TEST_APPLICATION_KEY_ID`.

b2sdk/account_info/sqlite_account_info.py
spartan218: I believe after our convention discussion in slack it will be 
`Return application key ID.`

b2sdk/account_info/in_memory.py
```diff
@@ -35,8 +35,8 @@ class InMemoryAccountInfo(UrlPoolAccountInfo):
     """
     Holder for all account-related information that needs to be kept
     between API calls, and between invocations of the command-line
-    tool.  This includes: account id, application key, auth tokens,
-    API URL, download URL, and uploads URLs.
+    tool.  This includes: account id, application key ID, application key,
```
ppolewicz:
```suggestion
    tool.  This includes: account ID, application key ID, application key,
```

test/test_account_info.py
```diff
@@ -280,7 +280,7 @@ def test_upgrade_1_default_allowed(self):
 
     def test_upgrade_2_default_app_key(self):
         """
-        The 'account_id_or_app_key_id' field should default to the account id.
+        The 'application_key_id' field should default to the account id.
```
ppolewicz:
```suggestion
        The 'application_key_id' field should default to the account ID.
```

b2sdk/raw_api.py
spartan218: @bwbeach should I make a check for either `TEST_ACCOUNT_ID` or `TEST_APPLICATION_KEY_ID`?

ppolewicz:
Coveralls has a false positive here due to `raw_api` not being covered by unit tests (we use the simulator in our unit tests). Actually `raw_api` is being tested by travis, but we don't tell coveralls about it (whether it was a deliberate decision to handle it this way or not, I think it might be a good decision, because those tests don't really do much and I'd rather keep coverage tracking meaningful for `b2http`).

ppolewicz:
continued in #28

PR #26, opened 2019-04-13
Better versioning

ppolewicz:
early PR to collect feedback


b2sdk/version.py
```diff
@@ -18,3 +23,178 @@
 PYTHON_VERSION = '.'.join(map(str, sys.version_info[:3]))  # something like: 2.7.11
 
 USER_AGENT = 'backblaze-b2/%s python/%s' % (VERSION, PYTHON_VERSION)
+
+
+class Deprecat:
```
bwbeach: I'm lost.  How about a doc string for the class.  What does it do?

bwbeach: It looks like you decorate a method with `Deprecat`, and it will issue warnings when you call it.

b2sdk/version.py
```diff
@@ -18,3 +23,178 @@
...
+    def _get_verified_versions(self, tuple_, subject, from_=None, allow_future=False):
+        """
+        subject may be "function", "method", "argument", "class" etc
+        returns None if replacement is going to happen in the future, or a 3-tuple with replacemnt specification:
```
spartan218: replacement

b2sdk/version.py
```diff
@@ -18,3 +23,178 @@
...
+            * name to replace to
+            * version where the change was introduced
+            * version where the support for the old name is going to be dropped
+        """
+        to_, version_changed, version_dropped = tuple_
+        assert len(version_changed) == len(version_dropped) == len(
```
bwbeach: Will fail with multi-digit numbers in versions, like `0.1.12`?

b2sdk/version.py
```diff
@@ -18,3 +23,178 @@
...
+            )
+        if version_changed >= VERSION:
+            return None
+        return to_, version_changed, version_dropped
+
+    def __init__(
+        self,
+        replacements=None,
+        rename=None,
+        allow_future_replacements=False,
```
bwbeach: It looks like `allow_future_*` lets you annotate deprecations before they actually happen.  Do we need that complexity?  Could we just add comments about future deprecations?

b2sdk/version.py
```diff
@@ -18,3 +23,178 @@
...
+        """
+        to_, version_changed, version_dropped = tuple_
+        assert len(version_changed) == len(version_dropped) == len(
+            VERSION
+        ) == 5  # TODO the comparator used below is too naiive for the provided version number
+        assert version_changed < version_dropped, '%s decorator is set to start renaming %s %r starting at version %s and finishing in %s. It needs to start at a lower version and finish at a higher version.' % (
+            self.__class__.__name__, subject, from_, version_changed, version_dropped
+        )
+        assert VERSION < version_dropped, '%s decorator is still used in version %s when old %s name %r was scheduled to be dropped in %s. It is time to remove that mapping.' % (
+            self.__class__.__name__, VERSION, subject, from_, version_dropped
```
spartan218: I might simplify this message to `, _subject is no longer in use as of _version_dropped.`

b2sdk/version.py
spartan218: just looked at the Semantic Version doc, technically there is no limit on digits on the version string.
Maybe we should just check that the versions are digits separated by `.`?

b2sdk/version.py
```diff
@@ -18,3 +23,178 @@
...
+        )
+        if not allow_future:
+            assert version_changed <= VERSION, '%s decorator indicates that the replacement of %s %s should take place in the future version %s, while the current version is %s' % (
```
spartan218: `decorator indicates that the replacement of %s %s ` should that that a to in between `subject` and `from_`

b2sdk/version.py
spartan218: I find the following helpful : 
1) the current method **is** deprecated
2) the current method **is going** to be deprecated as of x.x.x

b2sdk/version.py
ppolewicz: the only reasonable solution seems to be setuptools and `from pkg_resources import parse_version`

b2sdk/version.py
ppolewicz: it does not fully explain what happened

b2sdk/version.py
ppolewicz: I removed the `allow_future_*` flags

b2sdk/version.py
ppolewicz: all messages were changed

PR #28, opened 2019-04-19
App key renaming2

ppolewicz:
continuation of #25

### DIFF
--- a/b2sdk/account_info/abstract.py
+++ b/b2sdk/account_info/abstract.py
@@ -114,6 +114,7 @@ class AbstractAccountInfo(object):
         :rtype: str
         """
 
+    # TODO: make a decorator for get_account_id_or_app_key_id()
     @abstractmethod
     def get_application_key_id(self):
         """ 
@@ -258,6 +259,7 @@ class AbstractAccountInfo(object):
             ('capabilities' in allowed) and ('namePrefix' in allowed)
         )
 
+    # TODO: make a decorator for set_auth_data()
     @abstractmethod
     def _set_auth_data(
         self,
--- a/b2sdk/account_info/abstract.py
+++ b/b2sdk/account_info/abstract.py
@@ -117,7 +117,7 @@ class AbstractAccountInfo(object):
     # TODO: make a decorator for get_account_id_or_app_key_id()
     @abstractmethod
     def get_application_key_id(self):
-        """ 
+        """
         Returns the application key ID used to authenticate
 
         :rtype: str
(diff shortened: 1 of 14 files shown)


ppolewicz:
Example in quick_start guide is fixed on another branch

b2sdk/account_info/abstract.py
```diff
@@ -109,30 +110,42 @@ def clear_bucket_upload_data(self, bucket_id):
...
 
     @abstractmethod
+    def get_application_key_id(self):
+        """
+        Returns the application key ID used to authenticate
+
+        :rtype: str
+        """
+
+    @version_utils.rename_method(get_application_key_id, '0.1.5', '0.2.0')
```
spartan218: So clean, very nice.

spartan218: just curious do you have to include `get_account_id_or_app_key_id` the old method reference anywhere when doing this?

pre-commit.sh
```diff
@@ -133,7 +133,7 @@ done
 
 header test_raw_api
 
-if TEST_ACCOUNT_ID="$(head -n 1 ~/.b2_auth)" TEST_APPLICATION_KEY="$(tail -n 1 ~/.b2_auth)" python -m b2sdk.raw_api
+if TEST_APPLICATION_KEY_ID="$(head -n 1 ~/.b2_auth)" TEST_APPLICATION_KEY="$(tail -n 1 ~/.b2_auth)" python -m b2sdk.raw_api
```
spartan218: @bwbeach I think I asked this last time, do we need to include using `TEST_ACCOUNT_ID` somewhere?
Please show me how we build the python stuff we were talking about and I can add it.

spartan218:
I think we can close #25 ?

spartan218:
Last thing I thought of. I might write in the README.txt about how to deprecate methods or change param names so that other people can use the decorator you built. Thoughts?

pre-commit.sh
ppolewicz: since tests have passed in Travis on this PR, it must mean @bwbeach has already adjusted the CI settings

ppolewicz:
The current solution of method/parameter deprecation is provisional, so that we can make a release early next week. Rather than explaining in our readme how to deprecate, we should point to a documentation on deprecation that describes all of it, not just the decorator.

It is somewhat clean, but the target solution should be even more clean and it should either be moved to a separate project&repository, or the features that I've added should be added to one of the existing frameworks for deprecation. None of them have supported parameter renaming so it had to be written anyway, so that is the part we are shipping now.

bwbeach:
Are we ready to merge this?

ppolewicz:
Yes. Moreover I think we are ready to release 0.1.6.

PR #32, opened 2019-04-23
Fix truncated downloads not retrying

ppolewicz:


b2sdk/transferer/simple.py
```diff
@@ -22,10 +25,38 @@ def is_suitable(self, metadata, progress_listener):
...
             bytes_read += len(data)
+
+        # now normally bytes_read == metadata.content_length but sometimes there is a timeout
+        # or something and the server closes connection, while neither tcp or http have a problem
+        # with the truncated output, so we detect it here and try to continue
+
+        retries_left = self._num_tries - 1
+        while retries_left and bytes_read < metadata.content_length:
+            new_range = self._get_remote_range(response,
+                                               metadata).subrange(bytes_read, actual_size - 1)
```
bwbeach: `actual_size - 1` would be wrong if `actual_size` were 0, but I think that can't happen because the body of this loop should never run for empty files.

b2sdk/transferer/simple.py
```diff
@@ -22,10 +25,38 @@ def is_suitable(self, metadata, progress_listener):
...
+                new_range.as_tuple(),
+            ) as followup_response:
+                for data in followup_response.iter_content(
+                    chunk_size=self._get_chunk_size(actual_size)
+                ):
+                    file.write(data)
+                    digest.update(data)
+                    bytes_read += len(data)
+            retries_left -= 1
         return bytes_read, digest.hexdigest()
```
bwbeach: Is it worth double-checking that bytes_read == metadata.content_length?

If it's too small, maybe raise an exception?  If it's too big, that's a different problem, but probably worth raising an exception for.

b2sdk/transferer/simple.py
ppolewicz: indeed. It's worth documenting in an `assert` though!

b2sdk/transferer/simple.py
ppolewicz: It is checked by the caller, in a common place for all downloaders.

PR #33, opened 2019-04-23
Add timeout to python requests, handling backblaze API hangs

### LINKED ISSUE #31: Hang when backblaze endpoint stops responding
Several times over the past several days we have witnessed `b2 upload-file` hang.  We run these commands with `--noProgress` and there is no output that we can see from our logging that indicates that anything went wrong, however the command hangs for >24h until a human intervenes.  In debugging this we've seen a connection open to backblaze but no network traffic is being sent or received from the process.  Every time we kill and restart the process, it succeeds, so debugging this is is rather tricky.

We have experienced nearly identical issues with other python projects that were traced back to a lack of a `timeout` parameter in python `requests`.  I can't confirm this is exactly what is happening, but I do see what appears to be similar code issues with `.post` and `.get` which seem to be at the core of the b2 HTTP API (both before the 1.4.0 release and after):

https://github.com/Backblaze/b2-sdk-python/blob/master/b2sdk/b2http.py#L290
https://github.com/Backblaze/b2-sdk-python/blob/master/b2sdk/b2http.py#L358

Without a timeout there, if the server simply stops responding but keeps the socket open, my understanding and experience is that these calls will hang indefinitely.

From the [requests docs](https://2.python-requests.org//en/master/user/quickstart/#timeouts):

> You can tell Requests to stop waiting for a response after a given number of seconds with the timeout parameter. **Nearly all production code should use this parameter in nearly all requests**. Failure to do so can cause your program to hang indefinitely:

```
>>> requests.get('https://github.com/', timeout=0.001)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
requests.exceptions.Timeout: HTTPConnectionPool(host='github.com', port=80): Request timed out. (timeout=0.001)

```
> Note timeout is not a time limit on the entire response download; rather, an exception is raised if the server has not issued a response for timeout seconds (more precisely, if no bytes have been received on the underlying socket for timeout seconds). **If no timeout is specified explicitly, requests do not time out.**

jhart-r7:
Fixes #31.  Maybe.  I can't reproduce the original errors reliably enough.

### DIFF
--- a/b2sdk/b2http.py
+++ b/b2sdk/b2http.py
@@ -28,9 +28,6 @@ from six.moves import range
 
 logger = logging.getLogger(__name__)
 
-# the timeout used for requests
-TIMEOUT = 10
-
 def _print_exception(e, indent=''):
     """
     Used for debugging to print out nested exception structures.
@@ -238,6 +235,10 @@ class B2Http(object):
 
     """
 
+
+    # timeout for HTTP GET/POST requests
+    TIMEOUT = 120
+
     def __init__(self, requests_module=None, install_clock_skew_hook=True):
         """
         Initialize with a reference to the requests module, which makes
@@ -289,7 +290,7 @@ class B2Http(object):
         def do_post():
             data.seek(0)
             self._run_pre_request_hooks('POST', url, headers)
-            response = self.session.post(url, headers=headers, data=data, timeout=TIMEOUT)
+            response = self.session.post(url, headers=headers, data=data, timeout=self.TIMEOUT)
             self._run_post_request_hooks('POST', url, headers, response)
             return response
 
@@ -357,7 +358,7 @@ class B2Http(object):
         # Do the HTTP GET.
         def do_get():
             self._run_pre_request_hooks('GET', url, headers)
-            response = self.session.get(url, headers=headers, stream=True, timeout=TIMEOUT)
+            response = self.session.get(url, headers=headers, stream=True, timeout=self.TIMEOUT)
             self._run_post_request_hooks('GET', url, headers, response)
             return response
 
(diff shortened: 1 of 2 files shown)


ppolewicz:
We haven't seen that issue so far, B2 server used to have a ~120s timeout on its end which always resolved the problem. I agree that the client should have a timeout though, however:
1. I think it should be stored in `b2sdk.b2http.B2Http.TIMEOUT`, not in `b2sdk.b2http.TIMEOUT`
2. There are some corner cases in which the receiving device may loose NFS for a moment etc - setting the value to just 10 seconds, in my opinion, is likely to cause malfuction for some users. I think the default value should be larger than 120s (128s is the next "even" number). I don't know exactly why the server side used a 120s timeout but I bet there was some thought put into that decision by really smart engineers and so I'd rather follow that than try our luck with a small value.
3. `test.test_b2http.TestB2Http.test_get_content` has failed after your change. Please adjust the test to add the timeout parameter to the mock (it's a 1-line change). In the future you can execute `./pre-commit.sh` locally to run tests, some linters, code formatters etc.

b2sdk/b2http.py
```diff
@@ -236,6 +236,9 @@ class B2Http(object):
 
     """
 
+    # timeout for HTTP GET/POST requests
+    TIMEOUT = 120
```
ppolewicz: in case of a server-side issue, setting it at exactly 120s will cause the client to raise a timeout exception a few milliseconds before we receive a FIN packet from the server. It will cause all timeout errors to appear to be client-side and thus will suggest a client-side issue, while in reality most of those issues will actually be caused by the server. It would be nice for this to be easily recognizeable by the stacktrace in most of the cases, to only reserve the client-side timeout for a very misbehaving server.
Therefore please increase the timeout by at least 5 seconds (as heavily overloaded networks frequently have 4+s latency).

PR #35, opened 2019-04-23
Check return code from `make html` in Travis build.

bwbeach:


ppolewicz:
`cd ..` will only be executed if docs build passes, causing the rest of this CI run to be executed in a different workdir.
A possible alternative:
```
make html; status="$?"
cd ..
if [ "$status" -eq 0]; then
...
```

bwbeach:
Or how about using parens to make a sub-shell:

```
if (cd doc && make html); then
...
```

PR #43, opened 2019-06-03
Remove legacy ls

ppolewicz:


b2sdk/v0/api.py
```diff
@@ -0,0 +1,17 @@
...
+#
+# License https://www.backblaze.com/using_b2_code.html
+#
+######################################################################
+
+from .bucket import Bucket, BucketFactory
+from b2sdk.v1 import B2Api
+
+
+class B2Api(B2Api):
```
bwbeach: It might be clearer here to be explicit:

```
import b2sdk.v1

class B2Api(v1.B2Api):
```

b2sdk/v0/bucket.py
```diff
@@ -0,0 +1,18 @@
...
+# Copyright 2019 Backblaze Inc. All Rights Reserved.
+#
+# License https://www.backblaze.com/using_b2_code.html
+#
+######################################################################
+
+from b2sdk.v1 import Bucket, BucketFactory
+
+
+class Bucket(Bucket):
```
bwbeach: same comment here

PR #46, opened 2019-06-11
Restore testing on windows via Appveyor

ppolewicz:


ppolewicz:
@bwbeach please see https://github.com/Backblaze/b2-sdk-python/pull/46/files#diff-b6842e801063af34b2f7e507815a4f7dR343 - I'm not sure if it should be like that. On one hand it looks ugly, but on the other this is the right way of writing paths.

bwbeach:
Yes.  That's ugly. 

I don't understand [paths on Windows](https://blogs.msdn.microsoft.com/jeremykuhne/2016/04/21/path-normalization/).  I can ask the Windows people when they get in the office this morning.

PR #49, opened 2019-06-14
New sync API interface

ppolewicz:
documentation to be done after the new interface is accepted

### DIFF
--- a/b2sdk/sync/policy.py
+++ b/b2sdk/sync/policy.py
@@ -94,7 +94,12 @@ class AbstractFileSyncPolicy(object):
 
     @classmethod
     def files_are_different(
-            cls, source_file, dest_file, compare_threshold=None, compare_version_mode=None, newer_file_mode=None,
+            cls,
+            source_file,
+            dest_file,
+            compare_threshold=None,
+            compare_version_mode=None,
+            newer_file_mode=None,
     ):
         """
         Compare two files and determine if the the destination file
@@ -128,9 +133,8 @@ class AbstractFileSyncPolicy(object):
 
             logger.debug(
                 'File %s: source time %s, dest time %s, diff %s, threshold %s, diff > threshold %s',
-                source_file.name, source_mod_time, dest_mod_time, diff_mod_time, compare_threshold,
-                compare_threshold_exceeded
-            )
+                source_file.name, source_mod_time, dest_mod_time,
+                diff_mod_time, compare_threshold, compare_threshold_exceeded,)
 
             if compare_threshold_exceeded:
                 # Source is newer
@@ -144,9 +148,9 @@ class AbstractFileSyncPolicy(object):
                     elif newer_file_mode == cls.NEWER_FILE_SKIP:
                         return False
                     else:
-                        raise DestFileNewer(
-                            dest_file, source_file, cls.DESTINATION_PREFIX, cls.SOURCE_PREFIX
-                        )
+                        raise DestFileNewer(dest_file, source_file,
+                                            cls.DESTINATION_PREFIX,
+                                            cls.SOURCE_PREFIX)
 
         # Compare using file size
         elif compare_version_mode == cls.COMPARE_VERSION_SIZE:
@@ -158,9 +162,8 @@ class AbstractFileSyncPolicy(object):
 
             logger.debug(
                 'File %s: source size %s, dest size %s, diff %s, threshold %s, diff > threshold %s',
-                source_file.name, source_size, dest_size, diff_size, compare_threshold,
-                compare_threshold_exceeded
-            )
+                source_file.name, source_size, dest_size, diff_size,
+                compare_threshold, compare_threshold_exceeded,)
 
             # Replace if size difference is over threshold
             return compare_threshold_exceeded
@@ -237,13 +240,14 @@ class UpAndDeletePolicy(UpPolicy):
     """
 
     def _get_hide_delete_actions(self):
-        for action in super(UpAndDeletePolicy, self)._get_hide_delete_actions():
+        for action in super(UpAndDeletePolicy,
+                            self)._get_hide_delete_actions():
             yield action
         for action in make_b2_delete_actions(
-            self._source_file,
-            self._dest_file,
-            self._dest_folder,
-            self._transferred,
+                self._source_file,
+                self._dest_file,
+                self._dest_folder,
+                self._transferred,
         ):
             yield action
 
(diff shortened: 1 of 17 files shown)


b2sdk/sync/policy.py
```diff
@@ -29,7 +29,29 @@ class AbstractFileSyncPolicy(object):
     DESTINATION_PREFIX = NotImplemented
     SOURCE_PREFIX = NotImplemented
 
-    def __init__(self, source_file, source_folder, dest_file, dest_folder, now_millis, args):
+    NEWER_FILE_SKIP = 101
+    NEWER_FILE_REPLACE = 102
+
+    NEWER_FILE_MODES = [NEWER_FILE_REPLACE, NEWER_FILE_SKIP]
```
bwbeach: Is there no mode for "raise an error if dest file is newer"?

b2sdk/sync/policy.py
```diff
@@ -29,7 +29,29 @@ class AbstractFileSyncPolicy(object):
     DESTINATION_PREFIX = NotImplemented
     SOURCE_PREFIX = NotImplemented
 
-    def __init__(self, source_file, source_folder, dest_file, dest_folder, now_millis, args):
+    NEWER_FILE_SKIP = 101
```
bwbeach: Are the ints for speed?  Strings would make debugging easier.

b2sdk/sync/policy.py
```diff
@@ -65,10 +87,23 @@ def _should_transfer(self):
             return True
         else:
             # Both exist.  Transfer only if the two are different.
```
bwbeach: It's more complicated than "only if the two are different".  Not sure how much detail should be in this comment.

b2sdk/sync/policy.py
bwbeach: Oh.  Is `None` the other choice?

b2sdk/sync/sync.py
```diff
@@ -180,52 +96,107 @@ def count_files(local_folder, reporter):
...
+        :param dest_folder: destination folder object
+        :type dest_folder: b2sdk.sync.folder.AbstractFolder
+        :param now_millis: current time in milliseconds
+        :type now_millis: int
+        """
+        # For downloads, make sure that the target directory is there.
+        if dest_folder.folder_type() == 'local' and not self.dry_run:
+            dest_folder.ensure_present()
+
+        if source_folder.folder_type() == 'local' and not self.allow_empty_source:
```
bwbeach: I forget... Does `allow_empty_source` only apply to local sources, and not to syncing from B2 to local?

b2sdk/sync/sync.py
```diff
@@ -259,15 +231,113 @@ def sync_folders(
...
+        :type source_folder: b2sdk.sync.folder.AbstractFolder
+        :param dest_folder: destination folder object
+        :type dest_folder: b2sdk.sync.folder.AbstractFolder
+        :param now_millis: current time in milliseconds
+        :type now_millis: int
+        :param reporter: reporter object
+        :param policies_manager: policies manager object
+        """
+        if self.keep_days_or_delete == self.KEEP_DAYS_MODE and dest_folder.folder_type() == 'local':
+            raise InvalidArgument('keep_days_or_delete', 'cannot be used for local files')
```
bwbeach: It's only `KEEP_DAYS_MODE` that is not allowed.  Deleting is ok, right?

b2sdk/sync/sync.py
```diff
@@ -259,15 +231,113 @@ def sync_folders(
...
+            dest_file,
+            dest_folder,
+            now_millis,
+            delete,
+            keep_days,
+            self.newer_file_mode,
+            self.compare_threshold,
+            self.compare_version_mode,
+        )
+        for action in policy.get_all_actions():
```
bwbeach: Is there any reason this method should not return `policy.get_all_actions()`?

b2sdk/v0/sync.py
```diff
@@ -0,0 +1,189 @@
...
+from b2sdk.v1.exception import InvalidArgument, IncompleteSync
+from b2sdk.v1 import AbstractFileSyncPolicy
+from b2sdk.v1 import DEFAULT_SCAN_MANAGER
+from b2sdk.v1 import SyncReport
+from b2sdk.v1 import Synchronizer as SynchronizerV1
+
+logger = logging.getLogger(__name__)
+
+
+class Synchronizer(SynchronizerV1):
```
bwbeach: How about a class doc comment saying what this adapter is doing?

b2sdk/v0/sync.py
```diff
@@ -0,0 +1,189 @@
...
+            raise DestFileNewer(e.dest_file, e.source_file, e.dest_prefix, e.source_prefix)
+
+    def sync_folders(self, *args, **kwargs):
+        try:
+            super(Synchronizer, self).sync_folders(*args, **kwargs)
+        except IncompleteSync as e:
+            raise CommandError(e.message)
+
+
+def get_synchronizer_from_args(
```
bwbeach: Nice!

b2sdk/sync/policy.py
ppolewicz: looks like it. I'll fix it.

b2sdk/sync/policy.py
ppolewicz: We should be using enums, but 2.7 doesn't have them. ints are faster, but not that much. Django coders get used to using those due to the necessity of putting the enum in a db column, where speed matters way more than here. I'll fix it.

b2sdk/sync/policy.py
ppolewicz: this comment is very old and not part of this PR changes, but I agree, it can be misleading and needs to go.

b2sdk/sync/sync.py
ppolewicz: this code was just moved from one place to another, haven't thought about this in a while. It looks like it may be a bug or intended behavior, no idea. I'm fine with it as it is - if you think it is wrong, then I suggest that we discuss it in a separate issue.

b2sdk/v0/sync.py
ppolewicz: new style docs are tbd awaiting general interface approval. Now that we have it, I'll add docstrings, modify `doc/quick_start.rst` etc

b2sdk/sync/policy.py
bwbeach: Ok.

b2sdk/sync/sync.py
bwbeach: I agree that it's a separate issue.  It's probably always been this way.  (And I probably wrote it.)

ppolewicz:
We will add documentation and use enum/enum34 for various sync modes

ppolewicz:
Merged #52 instead

Issue #17, opened 2019-04-06
Improve sync_folders interface for library callers


ppolewicz:
```py
def sync_folders(
    source_folder,
    dest_folder,
    args,
    now_millis,
    stdout,
    no_progress,
    max_workers,
    policies_manager=DEFAULT_SCAN_MANAGER,
    dry_run=False,
    allow_empty_source=False,
):
```
`args` mandatory argument is a choice that we've made in the CLI times and an area which we can improve on.

mpnowacki-reef:
@ppolewicz I think this is resolved

ppolewicz:
yes

Issue #18, opened 2019-04-06
Remove raising CommandError in make_folder_sync_actions


ppolewicz:
```py
def make_folder_sync_actions(
    source_folder, dest_folder, args, now_millis, reporter, policies_manager=DEFAULT_SCAN_MANAGER
):
    """
    Yields a sequence of actions that will sync the destination
    folder to the source folder.
    """
    if args.skipNewer and args.replaceNewer:
        raise CommandError('--skipNewer and --replaceNewer are incompatible')

    if args.delete and (args.keepDays is not None):
        raise CommandError('--delete and --keepDays are incompatible')

    if (args.keepDays is not None) and (dest_folder.folder_type() == 'local'):
raise CommandError('--keepDays cannot be used for local files')
```
In sdk those should probably be assertions and we should have similar checks in the CLI with proper reporting to the user

spartan218:
@anuj-reef Was this issue solved in your PR? When I looked we were still raising CommandErrors.

mpnowacki-reef:
@ppolewicz I think this is resolved

ppolewicz:
so they are only raised in v0?

mpnowacki-reef:
in `v-latest` `CommandError` is only raised in `_parse_bucket_and_folder` and `LocalFolder.ensure_non_empty`. I will refactor both of them to use more appropriate errors.

PR #52, opened 2019-06-21
Improve Sync interface

anuj-reef:
This is to resolve issues #17 and #18.

### DIFF
--- a/b2sdk/sync/exception.py
+++ b/b2sdk/sync/exception.py
@@ -8,7 +8,7 @@
 #
 ######################################################################
 
-from ..exception import B2Error
+from ..exception import B2Error, B2SimpleError
 
 
 class EnvironmentEncodingError(B2Error):
@@ -35,3 +35,27 @@ export LANG=en_US.UTF-8""" % (
             self.filename,
             self.encoding,
         )
+
+
+class InvalidArgument(B2Error):
+    """
+    Raised when one or more arguments are invalid
+    """
+
+    def __init__(self, field_name, message):
+        """
+
+        :param field_name:
+        :param message:
+        :return:
+        """
+        super(InvalidArgument, self).__init__()
+        self.field_name = field_name
+        self.message = message
+
+    def __str__(self):
+        return "%s %s" % (self.field_name, self.message)
+
+
+class IncompleteSync(B2SimpleError):
+    pass
--- a/b2sdk/sync/exception.py
+++ b/b2sdk/sync/exception.py
@@ -41,17 +41,17 @@ class InvalidArgument(B2Error):
     Raised when one or more arguments are invalid
     """
 
-    def __init__(self, field_name, message):
+    def __init__(self, parameter_name, message):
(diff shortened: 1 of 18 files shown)


ppolewicz:
@bwbeach should `InvalidArgument` have child exceptions for every issue or should it be one class like now?

Also, is this an appropriate place to use `isinstance`? We did a tuple trick here, but perhaps `isinstance` here is better, despite our general approach to avoid it (in non-enum classes)?

b2sdk/sync/exception.py
```diff
@@ -35,3 +34,25 @@ def __str__(self):
             self.filename,
             self.encoding,
         )
+
+
+class InvalidArgument(B2Error):
```
bwbeach: I think that having a single `InvalidArgument` exception is fine.  Having one for each case seems like a lot of trouble, without much benefit.

b2sdk/sync/exception.py
```diff
@@ -35,3 +34,25 @@ def __str__(self):
...
+    """
+    Raised when one or more arguments are invalid
+    """
+
+    def __init__(self, field_name, message):
```
bwbeach: I think that technically, this should be `parameter_name`, not `field_name`.  A parameter is something passed into  a function, and a field is a slot in an object.

doc/source/api/sync.rst
```diff
@@ -18,6 +18,176 @@ The **high performance** of sync is credited to parallelization of:
...
+Sync Options
+============
 
-.. todo::
-   practically the whole sync documentation is missing
+Following are the important optional arguments that can be provided while initializing `Synchronizer` class.
+
+
+* ``compare_version_mode``: When comparing the source and destination files for finding whether to replace them or not, `compare_version_mode` can be passed to specify the mode of comparision. For possible values see :class:`b2sdk.v1.CompareVersionMode`. Default value is :py:attr:`b2sdk.v1.CompareVersionMode.MODTIME`
+* ``compare_threshold``: It's the minimum size/modification time difference between source and destination files before we assume that it is new and replace.
```
bwbeach: Might be worth specifying units here: bytes for size, and seconds for time.

doc/source/api/sync.rst
```diff
@@ -18,6 +18,176 @@ The **high performance** of sync is credited to parallelization of:
 Synchronizer spawns threads to perform the operations listed above in parallel to shorten
 the backup window to a minimum.
 
+Sync Options
```
bwbeach: Nice work on this section!

Having all of the examples will be very helpful.

b2sdk/sync/exception.py
ppolewicz: We can always extend it later on because `except` "uses" `isinstance` which works well with child classes...

But I shall add information about this to versioning policy so that public users of v1 know they must expect a more detailed exception - this will let us reduce the amount of proxies if we ever decide to make any changes here.

PR #53, opened 2019-06-24
make travis build more reliable

ppolewicz:


bwbeach:
Those changes look fine.  How does it improve the build time?

ppolewicz:
Actually cache slows the build down. Also the builds on travis have very variable times on the same configuration, so it's hard to say if something changed or not. I'm going to remove the caching change and leave the `travis_retry` for reliability.

PR #55, opened 2019-06-27
First version of b2_copy_file

ppolewicz:


b2sdk/bucket.py
```diff
@@ -750,6 +750,36 @@ def hide_file(self, file_name):
...
+        content_type=None,
+        file_info=None,
+        destination_bucket_id=None,
+    ):
+        """
+        Creates a new file in this bucket by (server-side) copying from an existing file.
+
+        :param str file_id: file ID of existing file
+        :param str new_file_name: file name of the new file
+        :param tuple[int,int],None bytes_range: start and end offsets, default is the entire file
```
bwbeach: It would be good to say whether the end is exclusive or inclusive.

(The fact that range headers are inclusive is annoying.)

Issue #23, opened 2019-04-06
Better documentation for B2Api.__init__


ppolewicz:
what the default values are for the parameters, which are required, why is it set up in this way and what is the intended usage

spartan218:
`def __init__(self, account_info=None, cache=None, raw_api=None, max_upload_workers=10):` 

That is the current signature, the documentation says briefly what instances they could be. I can take a shot at filling in the questions you asked and send a PR.

PR #67, opened 2019-07-24
Adding documentation per issue #23 to for B2Api.__init__

spartan218:
We thought it would be helpful to specify and answer some questions
about the constructors and parameters for the api.py such as..

Related to issue #23 

- what are the default values for the parameters?
- which ones are required?
- what is the intended usage?

### DIFF
--- a/b2sdk/api.py
+++ b/b2sdk/api.py
@@ -75,13 +75,34 @@ class B2Api(object):
         :param account_info: an instance of :class:`~b2sdk.account_info.upload_url_pool.UrlPoolAccountInfo`,
                              or any custom class derived from
                              :class:`~b2sdk.account_info.abstract.AbstractAccountInfo`
+
+                     NOT REQUIRED
+                     DEFAULT VALUE: None
+
+                     The account_info object is responsible for holding account related information
+                     between API calls. If one is provided, we will attempt to use that first.
+                     Otherwise one will be built using Sql lite called SqliteAccountInfo()
+
         :param cache: an instance of the one of the following classes:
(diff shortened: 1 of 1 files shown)


spartan218:
I could be wrong, but after looking at the code, it seems that none of the parameters are required. 

1) we build a SqliteAccountInfo for them if no account_info is provided
2) we either use the cache provided, build one from the SqliteAccountInfo or use a DummyCache that does nothing
3) We either use a given raw_api or create one from `B2RawApi()`
4) max_upload_workers is either given or always defaults to 10 (so I guess we could consider this required?) 

Let me know if you think my additions are helpful and clear.

b2sdk/api.py
```diff
@@ -75,13 +75,34 @@ def __init__(self, account_info=None, cache=None, raw_api=None, max_upload_worke
...
+                      and attempts to cache the data for future access.
+
         :param raw_api: an instance of one of the following classes:
                         :class:`~b2sdk.raw_api.B2RawApi`, :class:`~b2sdk.raw_simulator.RawSimulator`,
                         or any custom class derived from :class:`~b2sdk.raw_api.AbstractRawApi`
+
+                        NOT REQUIRED
+                        DEFAULT VALUE: None
+
+                        The HTTP layer that makes well-formed request to the B2 API.
```
ppolewicz: It is not very informative.
The true reason we provide `raw_api` parameter here is to make network-less unit testing simple by using RawSimulator in tests and RawApi in production. This docstring here should convey that intention and function to the user.

spartan218:
Thanks for the input, ill take a second stab.

PR #80, opened 2019-10-30
Fix Backblaze/B2_Command_Line_Tool#595 (memory leak during download)

ppolewicz:


ppolewicz:
pip stopped working on appveyor...

PR #81, opened 2019-10-30
Add prefix argument to Bucket.list_unfinished_large_files()

ppolewicz:
to address Backblaze/B2_Command_Line_Tool#591

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -10,4 +10,5 @@ SPLIT_PENALTY_AFTER_OPENING_BRACKET=0
 [nosetests]
 verbosity=2
 where=test
+#where=test/v1/test_bucket.py:TestListUnfinished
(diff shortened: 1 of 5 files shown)


setup.cfg
```diff
@@ -10,4 +10,5 @@ SPLIT_PENALTY_AFTER_OPENING_BRACKET=0
 [nosetests]
 verbosity=2
 where=test
+#where=test/v1/test_bucket.py:TestListUnfinished
```
bwbeach: Should this be here?

ppolewicz: yes, otherwise when I want to run a single test, I forget how to do it and spend time reinventing it over and over ;)

bwbeach: Cool.

Issue #77, opened 2019-10-09
Transferer v2


ppolewicz:
We need to significantly alter how uploads are handed in b2sdk to provide better configurability and to properly handle multi-part server side copy.

The logic of uploads is going to be moved to transferer, but it is going to be pretty abstract. The same flow will be used by single part uploads, multi-part uploads, multi-part upload continuation, single-part server side copy, mutli-part server side copy as well as "emerging" - synthesyzing a file out of pieces, some of which are in the cloud already and some are still local and need to be uploaded. The Emerger will be capable of uploading, server-side copying and downloading a stream to then upload it, that's why it is not called an Uploader.

A corner case of that is when we need to download some data to then upload it again (due to server-side lower size limit of part equal currently to 5MB) and another is that sometimes a very large file needs to be copied and Emerge will need to break it down to smaller chunks, otherwise the server-side copy will refuse to handle the request.

The copy operation will have an optional flag, which will terminate the operation if any data would need to be downloaded in order to fulfill the copying (so if it would not be a purely server-side copy). Setting that flag to `True` will force the Emerger to consume the entire iterable to verify whether the operation is a pure copy (unless it is not - then it can exit as soon as a range that cannot be satisfied without downloading is found).

There will be two thread pools, one for uploading and another one for downloading. This is to avoid deadlocks when we'd need to download data in order to upload it back and all threads in the pool are busy doing that (waiting for someone to download the data). Since downloads will never wait for uploads (in the current design), two threadpools will ensure we avoid a deadlock.

To decrease the final PR size, we'll move the upload logic as is to transferer first. We'll implement b2_copy_part support in raw_api as well as in the simulator. Some tests can maybe be added before the feature is actually implemented.

The part of code which decides how to split a file into parts is going to have a simple implementation - it is a very hard problem to optimize and we are not going to spend a lot of effort on the strategy at this point. This may cause some copy operations which could be completed with just upload+copy to not be properly recognized, so if the `forbid_downloads` flag is set to `True`, we may return failure because of an imperfect strategy. This is a known limitation of the design and we will be open for pull requests from anyone wishing to optimize this further (as long as it doesn't degrade the performance of the naiive approach (too much)).

In Transferer we already delegate some functionality away to Downloaders. This is nice because Transferer class itself is smaller. Here the functionality will be moved to a new class, called Emerger. Emerger, in its main operation, will need to accept an iterable of RangeToEmerge objects, which may be:
[...]

PR #84, opened 2019-11-14
Refactor upload token logic to B2Session

anuj-reef:
As part of #77 we are moving upload url and token logic to B2Session where it actually belongs.

### DIFF
--- a/b2sdk/raw_api.py
+++ b/b2sdk/raw_api.py
@@ -60,8 +60,8 @@ class TokenType(Enum):
 
 def set_token_type(token_type):
     """
-    This is a decorator to identify the type of token to be used in session.
-    When the raw_api is used through B2Session, it will be used to identify the type of url and token to be used.
+    This is a decorator to identify the type of token that must be passed into a function.
+    When the raw_api is used through B2Session, it will be used to identify the type of url and token to be passed.
 
     :param token_type: TokenType enum
     :return:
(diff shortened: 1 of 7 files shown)


b2sdk/raw_api.py
```diff
@@ -51,6 +51,29 @@
...
+@unique
+class TokenType(Enum):
+    API = 'api'
+    UPLOAD_PART = 'upload_part'
+    UPLOAD_SMALL = 'upload_small'
+
+
+def set_token_type(token_type):
+    """
+    This is a decorator to identify the type of token to be used in session.
```
bwbeach: I would say this the other way, because this is the raw API layer, which should not be aware of the session layer above.

"This is a decorator to identify the type of token that must be passed into a function."

b2sdk/session.py
```diff
@@ -30,13 +30,22 @@ def __getattr__(self, name):
         @functools.wraps(f)
         def wrapper(*args, **kwargs):
             auth_failure_encountered = False
+            # A *magic* that will identify and generate the correct type of Url and token based on the decorator on the B2RawApi method.
+            token_type = getattr(f, 'token_type', TokenType.API)
```
bwbeach: I think I might put the knowledge of "what is the default token type" in the raw API layer, next to the token type decorator.  Maybe add a `get_token_type` function there that does this `getattr`.

ppolewicz: great idea, didn't think of that!

PR #85, opened 2019-11-18
B2Api documentation extension

magarwal-reef:


ppolewicz:
This is a followup of #67 and #23

Issue #79, opened 2019-10-29
400 bad_request more than one upload using auth token


fred-aero:
I have a loop uploading files locally to B2 cloud. After a few thousand uploads over a few hours, I get this error:
`b2sdk.exception.UnknownError: Unknown error: 400 bad_request more than one upload using auth token 4_0021a87abca5e6b0000000005_018fe4f8_97f3ec_uplg_wANv-9n9ent4nLDoDI0yMFmZqeQ=`

I feel like upload_local_file method should be able to handle renegotiating a new auth token on its own rather than failing as such.

ppolewicz:
upload_local_file does renegotiate an auth token if one expires. This is different.

First of all, usually it is better to `sync` to upload multiple files, than loop and `upload_local_file`, you should look into it. Nevertheless, a bug of some sort was observed and I am not sure what to think of it. Perhaps an upload request has failed/timed out, but was still alive on the server and we reused an **upload token** (not an **auth token**, as misleadingly hinted by the server) to upload another file? @bwbeach can you shed some light on why the server forbids uploading multiple files with one **auth token**?

ppolewicz:
The bug is here: https://github.com/Backblaze/b2-sdk-python/blob/ec0de1cc783a1f5caf8b4989b53eb7e1b7a7e766/b2sdk/bucket.py#L702

ppolewicz:
@fred-aero please install a development version from https://github.com/Backblaze/b2-sdk-python/archive/fix_79.zip (https://github.com/Backblaze/b2-sdk-python/tree/fix_79) with `python setup.py install` and let me know if it fixes the problem.

fred-aero:
I will for sure give it a shot, thanks!

ppolewicz:
@fred-aero if you could confirm that it works as intended, I'd like to ship that fix before working on some other larger changes in the same area of code. Could you test it soon?

fred-aero:
Sorry for the delay, I just set up two servers to perform a similar job to what gave me the error before. One server is using the posted fix while the other is acting as a control. I'm going to have them both run for a while and let you know what happens.

fred-aero:
The server running fix_79 encountered the same error unfortunately.
Here is the full stack trace:
```
Traceback (most recent call last):
  File "b2_stress_test.py", line 106, in <module>
    main()
  File "b2_stress_test.py", line 91, in main
    b2_upload(bucket_recv, local_name, object_name)
[...]

PR #89, opened 2019-11-30
Guard against concurrent usage of upload tokens

vmax:
* Guards simulator against #79
* Raises `UploadTokenUsedConcurrently` in case same error occurs


ppolewicz:
This is just the simulator change, the actual test to expose the issue will come in a separate PR

b2sdk/raw_simulator.py
```diff
@@ -654,6 +656,10 @@ def __init__(self):
         # Set of auth tokens that have expired
         self.expired_auth_tokens = set()
 
+        # Map from auth token to a lock that upload procedure acquires
+        # when utilizing the token
+        self.currently_used_auth_tokens = collections.defaultdict(threading.Lock)
```
bwbeach: Is there a danger that this mapping will grow unbounded?  If the the usage pattern creates a new upload token for each upload, every upload would add to the map.  When uploading millions of files it could get big.

Would it be sufficient just to keep a set of the currently used tokens?  (With locking for thread safety, of course.)

b2sdk/utils.py
```diff
@@ -366,6 +366,24 @@ class B2TraceMetaAbstract(DefaultTraceAbstractMeta):
     pass
 
 
+class ConcurrentUsedAuthTokenGuard(object):
```
bwbeach: It's nice to have a doc string for each class, with enough information to know how to use the class without reading the code.

b2sdk/raw_simulator.py
ppolewicz: This change is just for the raw_simulator, which is only used for unit tests in place of the actual cloud service. The problem here is that user reported a bug that clearly shows the CLI did not prevent token re-usage. Before we will try to implement an exposing test, we want to make the simulated cloud service raise alarm whenever someone is trying to reuse the token.

This whole simulator keeps all files in memory etc. I wouldn't worry about growth of the mapping in mentioned line - file content will consume all the memory first.

bwbeach: Ah.  Right.  Sounds good then.

b2sdk/utils.py
vmax: @bwbeach addressed in https://github.com/Backblaze/b2-sdk-python/pull/89/commits/7ec6e2589bdb48dbeda0661335727b502c3943bc

ppolewicz:
It looks like pypy-5.1 has found an interlace where the guard was hit and raised an exception, though it passed the tests previously. We cannot merge it yet as it will fail tests randomly.

ppolewicz:
This one is fun! It sporadically appears after 10-12 reruns only on PyPy - but only if you run the whole test suite; if you limit it to file that contains failing tests, it can succeeded ~100 times in a row). After ~10 sets of reruns it stopped appearing even when ran on PyPy. It never appears on CPython.

ppolewicz:
This is starting to be a heisenbug, so we will try to make a test which increases the probability of exposing the issue. If that won't work, we will try to add logs and tweak the cpython thread switching frequency knobs to make it crash on cpython in hopes of increasing reliability to a level which will let us fix it in a few steps of adding more and more logs until we will find the issue.

vmax:
@bwbeach @ppolewicz
Is this assumption (*) about the real B2 server correct: subsequent calls to `b2_get_upload_part_url` return *different* `authorizationToken`s every time? If so, this assumption isn't really held in `raw_simulator`, which always gives *the same* `authorizationToken` for the same `fileId`, see [ref](https://github.com/Backblaze/b2-sdk-python/blob/master/b2sdk/raw_simulator.py#L403-L405).

Not holding this assumption might've resulted in hitting the guard. Consider the following scenario:
- we upload a very large file (I used `self.simulator.MIN_PART_SIZE * 100`) which is split in chunks
- every chunk in uploaded on a different thread
- every thread requests a URL from the pool and uses it to do the upload (1)
- when pool is exhausted, a new URL is requested (2)
- URLs in (1) and (2), due to assumption (*) not being held, are equal
- so the guard is hit when using (2)

This is the test I've used:

```
    def test_upload_very_large(self):
        x = self._make_data(self.simulator.MIN_PART_SIZE * 100)
        self.bucket.upload_bytes(x, 'file1')
```

b2sdk/raw_simulator.py
```diff
@@ -402,7 +403,7 @@ def get_upload_url(self):
         return dict(bucketId=self.bucket_id, uploadUrl=upload_url, authorizationToken=upload_url)
 
     def get_upload_part_url(self, file_id):
-        upload_url = 'https://upload.example.com/part/%s' % (file_id,)
+        upload_url = 'https://upload.example.com/part/%s/%d' % (file_id, random.randint(1, 10**9))
```
ppolewicz: this will randomly cause a failure every 10 million executions or so. I'll fix it.

b2sdk/exception.py
```diff
@@ -438,6 +452,11 @@ def interpret_b2_error(status, code, message, response_headers, post_params=None
         return MissingPart(post_params.get('fileId'))
     elif status == 400 and code == "part_sha1_mismatch":
         return PartSha1Mismatch(post_params.get('fileId'))
+    elif status == 400 and code == "bad_request":
+        matcher = UPLOAD_TOKEN_USED_CONCURRENTLY_ERROR_MESSAGE_RE.match(message)
+        if matcher is not None:
```
ppolewicz: if matcher _is_ None, then a 400 `bad_request` error is going to be silently swallowed by that `if` branch!

b2sdk/raw_simulator.py
ppolewicz: actually more, because some tests may upload multiple files concurrently increasing the collision frequency

b2sdk/raw_simulator.py
```diff
@@ -31,9 +33,9 @@
     Unauthorized,
     UnsatisfiableRange,
 )
+from .raw_api import AbstractRawApi, HEX_DIGITS_AT_END, MetadataDirectiveMode, set_token_type, TokenType
```
ppolewicz, line 36:
```suggestion
from .raw_api import AbstractRawApi, HEX_DIGITS_AT_END, MetadataDirectiveMode, TokenType
```
sorry, bad merge

PR #90, opened 2019-12-05
Add bucket options strings set

keithele:
This update includes the new "options" strings set returned by the B2 API in the Bucket object.

### DIFF
--- a/test/v0/test_api.py
+++ b/test/v0/test_api.py
@@ -38,6 +38,7 @@ class TestApi(TestBase):
             'bucketInfo': {},
             'corsRules': [],
             'lifecycleRules': [],
+            'options': set(),
             'revision': 1,
         }
         assert delete_output == expected, delete_output
--- a/b2sdk/bucket.py
+++ b/b2sdk/bucket.py
@@ -138,6 +138,7 @@ class Bucket(object):
         lifecycle_rules=None,
         revision=None,
         bucket_dict=None,
+        options_set=None
     ):
         """
         :param b2sdk.v1.B2Api api: an API object
@@ -149,6 +150,7 @@ class Bucket(object):
         :param dict lifecycle_rules: lifecycle rules to store with a bucket
         :param int revision: a bucket revision number
         :param dict bucket_dict: a dictionary which contains bucket parameters
+        :param set options_set: set of bucket options strings
         """
         self.api = api
         self.id_ = id_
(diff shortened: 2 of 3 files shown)


ppolewicz:
Options don't seem to be documented in the API docs (yet?)

keithele:
@ppolewicz Ah, you're right.  It's not in the B2 API documentation yet.  It is being returned by the API in production right now, so wanted to update the SDK to handle it.  I've made a note to update the docs.  Thanks!

ppolewicz:
My point is that without the docs an additional field cannot really be used as nobody knows how to interpret the returned data. Is the resulting option set made of a limited number of strings that have some additional meaning? Because if that is true, we should parse them into a series of classes with properties that describe the behavior of the options to the SDK user to interpret them.

Can we wait with merging this until it is documented, so that your patch can be cross-reviewed?

keithele:
I've updated the B2 API documentation.  The `options` parameter doesn't have any values for now.  When strings are added in the future, then the possible values can be documented and added for use in the SDK.  But for now, there aren't any values to document, it's a future-looking/flexible set of strings.  Does that help?

ppolewicz:
Can you give an example of how it might be used?

keithele:
I don't have an example because there aren't any uses of it.  Just matching the current API.

test/v0/test_api.py
```diff
@@ -38,6 +38,7 @@ def test_list_buckets(self):
             'bucketInfo': {},
             'corsRules': [],
             'lifecycleRules': [],
+            'options': set(),
```
ppolewicz: This change should be done in `test/v1/test_api.py:test_list_buckets` too

keithele: Unfortunately, the `delete_bucket` api being used by `test/v1/test_api.py:test_list_buckets` doesn't have a return value/it's `None`.  So the existing test case is valid according to the existing API in `/api.py`.

Would we want to change `/api.py:delete_bucket()` to return the dictionary from `/raw_api.py:delete_bucket()`?  
Was there a reason the old API returned the value from `delete_bucket` and the new one doesn't?  What do you think?

ppolewicz: Oh, you are right!

I think it was listed in a comment as something that is not needed. Should the user of `delete_bucket()` expect to receive a dict/json in return to his (successful) call? Why?

keithele: Looks like you updated it back in June.  I think it works fine as-is.  If you're good with it we can mark this resolved and merge.

PR #91, opened 2019-12-17
B2session refactor

mzukowski-reef:
Make the B2Session code clearer and be explicit about each call rather than relying on magic (and even more magic after recent changes - this was not an ideal direction). Now the constructor of B2Session has defaults, which is also nice.

Furthermore, RawApi interface was repaired.

The use of magic in B2Session has been frowned upon for a very long time now and the review of recent changes (that made it even more magic) made us realize that it is time to restructure it properly.

### DIFF
--- a/b2sdk/api.py
+++ b/b2sdk/api.py
@@ -109,6 +109,11 @@ class B2Api(object):
 
     @property
     def raw_api(self):
+        """
+        .. warning::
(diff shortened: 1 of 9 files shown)


ppolewicz:
Another PR may follow this one to adjust the docs (now that the RawApi interface was repaired we can move docstrings from RawApi to AbstractRawApi)

PR #92, opened 2019-12-17
Pin coverage version to last known working with parallel executions

mzukowski-reef:


ppolewicz:
Coverage 5.0 broke parallel tests with sqlite

PR #100, opened 2020-01-13
Resolves Backblaze#99

magarwal-reef:


ppolewicz:
@magarwal-reef please rebase on current master to fix CI issues

PR #101, opened 2020-01-23
Fix CI

mzukowski-reef:


ppolewicz:
We will re-enable the license header checking after merging a large PR (which would have many conflicts due to a sizeable amount of code being moved)

PR #106, opened 2020-02-11
Fix 400 bad_request more than one upload using auth token

ppolewicz:
I hope it fixes the entire #79 now


ppolewicz:
rebased on master to avoid 3.4 CI failures

ppolewicz:
needs to be rebased again after #118 is merged

PR #110, opened 2020-03-05
Fix constructing Range from HTTP request Range header

mzukowski-reef:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Range_requests
```
GET /z4d4kWk.jpg HTTP/1.1
Host: i.imgur.com
Range: bytes=0-1023
```
```
HTTP/1.1 206 Partial Content
Content-Range: bytes 0-1023/146515
Content-Length: 1024
...
(binary content)
```

### DIFF
--- a/b2sdk/transferer/range.py
+++ b/b2sdk/transferer/range.py
@@ -24,9 +24,9 @@ class Range(object):
         """
         Factory method which returns an object constructed from Range http header.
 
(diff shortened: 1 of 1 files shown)


mzukowski-reef:
Correct HTTP Range header is `bytes=0-11` not `bytes 0-11`

PR #111, opened 2020-03-19
Fix b2cli 608

magarwal-reef:
Fixes https://github.com/Backblaze/B2_Command_Line_Tool/issues/608


ppolewicz:
This was fixed in #118, now we just add a regression test

PR #113, opened 2020-04-06
Emerger implementation

### LINKED ISSUE #77: Transferer v2

ppolewicz:
Resolves #77


bwbeach:
I read through the planner and the executor.  They look good.  

I didn't get deep enough to find things like off-by-one errors and such, and I'm assuming that the tests would cover those things.

ppolewicz:
Closing in favor of #118

PR #120, opened 2020-04-26
Add a `list_file_versions` API to buckets. Fixes #117.

### LINKED ISSUE #117: No public API for getting a list of file ids for a filename
In order to delete a file you need to know the file id of the version to delete but unless I'm missing something there is no straightforward way to list the different versions of a file with a given filename. `bucket.ls` exists but since it enforces a trailing `/` on the folder you pass the only way I can see right now to get this through the public API is to do something like this:

```python
def get_file_ids(bucket, filename):
    for file_info, folder_name in bucket.ls(folder_to_list=os.path.dirname(filename), show_versions=True):
        if file_info.file_name == filename:
            yield file_info
```

Obviously this potentially wastes bandwidth by returning all files in the same "folder".

Mossop:
Adds the API discussed in #117.

I ended up having to bypass the pre-commit.sh commit hook, it was failing the pyflakes test with a bunch of issues in files I haven't touched. Those same issues show up when running the tests on current master.

### DIFF
--- a/b2sdk/bucket.py
+++ b/b2sdk/bucket.py
@@ -207,11 +207,12 @@ class Bucket(object):
         if fetch_count is not None and fetch_count <= 0:
             # fetch_count equal to 0 means "use API default", which we don't want to support here
             raise ValueError("unsupported fetch_count value")
+        start_file_name = file_name
         start_file_id = None
         session = self.api.session
         while 1:
             response = session.list_file_versions(
-                self.id_, file_name, start_file_id, fetch_count, file_name
+                self.id_, start_file_name, start_file_id, fetch_count, file_name
             )
 
             for entry in response['files']:
@@ -220,10 +221,10 @@ class Bucket(object):
                     # All versions for the requested file name have been listed.
                     return
                 yield file_version_info
-            if response['nextFileName'] != file_name:
-                # Unless there are more files with the same name we can stop here.
-                return
+            start_file_name = response['nextFileName']
             start_file_id = response['nextFileId']
+            if start_file_name is None:
+                return
 
     def ls(self, folder_to_list='', show_versions=False, recursive=False, fetch_count=None):
         """
--- a/b2sdk/bucket.py
+++ b/b2sdk/bucket.py
@@ -196,6 +196,35 @@ class Bucket(object):
         """
         return self.api.list_parts(file_id, start_part_number, batch_size)
 
+    def list_file_versions(self, file_name, fetch_count=None):
+        """
+        Lists all of the versions for a single file.
+
+        :param str file_name: the name of the file to list.
+        :param int,None fetch_count: how many entries to list per API call or ``None`` to use the default. Acceptable values: 1 - 10000
+        :rtype: generator[b2sdk.v1.FileVersionInfo]
+        """
+        if fetch_count is not None and fetch_count <= 0:
+            # fetch_count equal to 0 means "use API default", which we don't want to support here
+            raise ValueError("unsupported fetch_count value")
+        start_file_id = None
+        session = self.api.session
+        while 1:
+            response = session.list_file_versions(
+                self.id_, file_name, start_file_id, fetch_count, file_name
+            )
+
+            for entry in response['files']:
+                file_version_info = FileVersionInfoFactory.from_api_response(entry)
+                if file_version_info.file_name != file_name:
+                    # All versions for the requested file name have been listed.
+                    return
+                yield file_version_info
+            if response['nextFileName'] != file_name:
+                # Unless there are more files with the same name we can stop here.
+                return
+            start_file_id = response['nextFileId']
+
     def ls(self, folder_to_list='', show_versions=False, recursive=False, fetch_count=None):
         """
         Pretend that folders exist and yields the information about the files in a folder.
--- a/test/v1/test_bucket.py
+++ b/test/v1/test_bucket.py
@@ -357,17 +357,17 @@ class TestListVersions(TestCaseWithBucket):
         ]
         self.assertEqual(expected, actual)
 
-    # def test_no_second_response(self):
-    #     data = six.b('hello world')
-    #     file_id = self.bucket.upload_bytes(data, 'a/b').id_
-    #     self.bucket.upload_bytes(data, 'a/b/c')
+    def test_all_versions_in_response(self):
+        data = six.b('hello world')
+        file_id = self.bucket.upload_bytes(data, 'a/b').id_
+        self.bucket.upload_bytes(data, 'a/b/c')
 
-    #     expected = [(file_id, 'a/b', 11, 'upload')]
-    #     actual = [
-    #         (info.id_, info.file_name, info.size, info.action)
-    #         for info in self.bucket.list_file_versions('a/b', fetch_count=1)
-    #     ]
-    #     self.assertEqual(expected, actual)
+        expected = [(file_id, 'a/b', 11, 'upload')]
+        actual = [
+            (info.id_, info.file_name, info.size, info.action)
+            for info in self.bucket.list_file_versions('a/b', fetch_count=1)
+        ]
+        self.assertEqual(expected, actual)
 
     def test_bad_fetch_count(self):
         try:
--- a/test/v1/test_bucket.py
+++ b/test/v1/test_bucket.py
@@ -303,6 +303,81 @@ class TestLs(TestCaseWithBucket):
         self.assertBucketContents(expected, '', show_versions=True)
 
 
+class TestListVersions(TestCaseWithBucket):
+    def test_single_version(self):
+        data = six.b('hello world')
+        a_id = self.bucket.upload_bytes(data, 'a').id_
+        b_id = self.bucket.upload_bytes(data, 'b').id_
+        c_id = self.bucket.upload_bytes(data, 'c').id_
+
+        expected = [(a_id, 'a', 11, 'upload')]
+        actual = [
+            (info.id_, info.file_name, info.size, info.action)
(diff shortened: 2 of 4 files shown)


Mossop:
I guess I misread the docstring issue, looks like it isn't in my code?
The python 3.4 failure also looks unlikely to be as a result of this change as I haven't modified any dependencies.

b2sdk/bucket.py
```diff
@@ -292,6 +292,36 @@ def list_parts(self, file_id, start_part_number=None, batch_size=None):
         """
         return self.api.list_parts(file_id, start_part_number, batch_size)
 
+    def list_file_versions(self, file_name, fetch_count=None):
+        """
+        Lists all of the versions for a single file.
+
+        Depending on a Bucket's settings many versions may exist for the same filename.
```
ppolewicz, line 299: As far as I know, there is no bucket setting in B2 that would disable this capability.

b2sdk/bucket.py
```diff
@@ -292,6 +292,36 @@ def list_parts(self, file_id, start_part_number=None, batch_size=None):
...
+        This lists all of the versions present in the Bucket.
+
+        :param str file_name: the name of the file to list.
+        :param int,None fetch_count: how many entries to list per API call or ``None`` to use the default. Acceptable values: 1 - 1000
```
ppolewicz, line 303:
```suggestion
        :param int,None fetch_count: how many entries to list per API call or ``None`` to use the default. Acceptable values: 1 - 10000
```

b2sdk/bucket.py
```diff
@@ -292,6 +292,36 @@ def list_parts(self, file_id, start_part_number=None, batch_size=None):
...
+        :rtype: generator[b2sdk.v1.FileVersionInfo]
+        :returns: generator of file_version_info
```
ppolewicz, line 305: I don't think we have to say it, the type makes it pretty clear already

b2sdk/bucket.py
```diff
@@ -292,6 +292,36 @@ def list_parts(self, file_id, start_part_number=None, batch_size=None):
...
+        """
+        start_file_id = None
+        session = self.api.session
+        while True:
```
ppolewicz, line 309:
```suggestion
        while 1:
```
since we still support python 2.7. It's a long story :)

b2sdk/bucket.py
```diff
@@ -292,6 +292,36 @@ def list_parts(self, file_id, start_part_number=None, batch_size=None):
...
+
+        Depending on a Bucket's settings many versions may exist for the same filename.
+        This lists all of the versions present in the Bucket.
+
+        :param str file_name: the name of the file to list.
+        :param int,None fetch_count: how many entries to list per API call or ``None`` to use the default. Acceptable values: 1 - 1000
+        :rtype: generator[b2sdk.v1.FileVersionInfo]
+        :returns: generator of file_version_info
+        """
+        start_file_id = None
```
ppolewicz, line 307:
```suggestion
        """
        if fetch_count is not None and fetch_count <= 0:
            # fetch_count equal to 0 means "use API default", which we don't want to support here
            raise ValueError("unsupported fetch_count value")
        start_file_id = None
```

b2sdk/bucket.py
```diff
@@ -292,6 +292,36 @@ def list_parts(self, file_id, start_part_number=None, batch_size=None):
...
+            )
+
+            for entry in response['files']:
+                file_version_info = FileVersionInfoFactory.from_api_response(entry)
+                if file_version_info.file_name != file_name:
+                    # All versions for the requested file name have been listed.
+                    return
+                yield file_version_info
+            if response['nextFileName'] != file_name:
+                # Unless there are more files with the same name we can stop here.
```
ppolewicz, line 321: a test to hit that condition is missing

b2sdk/bucket.py
Mossop: A bucket's lifecycle settings control how many versions of a file to keep, one of the options is to only keep the most recent version, though admittedly that does still mean more than one version can exist, as I understand it older versions get removed periodically.

b2sdk/bucket.py
Mossop: Interesting. I see that the ls function also uses `while True:`, is that an issue too?

b2sdk/bucket.py
Mossop: I'll change the wording a little to see if it makes more sense.

Mossop: I guess this just overcomplicates matters, I'll just remove it.

b2sdk/bucket.py
Mossop: I've pushed the latest patch rebased to master but the test I wrote for this case hit some issues so I've commented it out. I would like some advice on how to proceed.

In order to test this condition the number of versions of a file must match the fetch count and the expectation would be that `nextFileName` would then be the name of the next file (or at least something that sorts later than the current file). However the RawSimulator doesn't behave this way, the `nextFileName` it returns is always the same name as the last file returned in the results:
https://github.com/Backblaze/b2-sdk-python/blob/6d35959ccfb06c3888702d91f3ce8d699f24f8e0/b2sdk/raw_simulator.py#L520

On the surface that is harmless enough, it just makes testing that case difficult. However the nextFileId that RawSimulator returns is also problematic. In the test I wrote I upload a single file which gets the id '9999' so with a fetch_count of 1 the first request returns the only version and then RawSimulator returns a nextFileName matching the file name and a nextFileId of '10000': https://github.com/Backblaze/b2-sdk-python/blob/6d35959ccfb06c3888702d91f3ce8d699f24f8e0/b2sdk/raw_simulator.py#L521

So the next request we send the same start_file_name and a start_file_id of '10000'. As RawSimulator loops over its known files it finds our file with id 9999 but the IDs are strings and it uses a string comparison to decide whether to include the file ID or not: https://github.com/Backblaze/b2-sdk-python/blob/6d35959ccfb06c3888702d91f3ce8d699f24f8e0/b2sdk/raw_simulator.py#L514

The string '10000' sorts lower than the string '9999' so the second request returns the same file id as the first and the same nextFileName and nextFileId. So the current implementation of the new API would loop forever in this case.

I can try and patch the RawSimulator to resolve these issues but I'm not sure if it would make more sense to do that as a separate PR or how you'd prefer to proceed.

b2sdk/bucket.py
ppolewicz: It is not something I would go back and change but when I am reviewing newly added code, I flag this as a potential inefficiency. Fortunately we are not going to support 2.7 forever, so this should not be a problem for long.

b2sdk/bucket.py
ppolewicz: 👍

b2sdk/bucket.py
ppolewicz: If RawSimulator has a bug that disturbs your test, then we should fix it. I think extending/fixing it in a PR that needs the fix is appropriate.

The question is, how does the cloud server behave in this exact case? From what the b2 server docs say and from what you say, RawSimulator implementation is flawed, but before modifying it, please test the same case with the cloud server - I want to avoid a situation where the inconsistency is caused by an error in documentation of the cloud server while actually the behavior of RawSimulator and cloud server are the same.

Mossop: Good catch. Looks like the b2 server does return the same nextFileName as the last returned file, just with a later (and in this case non-existent) nextFileId. I'll update to work with that case.

test/v1/test_bucket.py
```diff
@@ -357,17 +357,17 @@ def test_ignores_subdirectory(self):
...
-    #     expected = [(file_id, 'a/b', 11, 'upload')]
-    #     actual = [
-    #         (info.id_, info.file_name, info.size, info.action)
-    #         for info in self.bucket.list_file_versions('a/b', fetch_count=1)
-    #     ]
-    #     self.assertEqual(expected, actual)
+    def test_all_versions_in_response(self):
+        data = six.b('hello world')
+        file_id = self.bucket.upload_bytes(data, 'a/b').id_
+        self.bucket.upload_bytes(data, 'a/b/c')
```
ppolewicz, line 363:
```suggestion
        self.bucket.upload_bytes(data, 'a/b')
```
without the suggested change, this will be skipped as in the test above, due to the file being a subdirectory, not due to the fetch_count limit being equal to 1

Mossop: If we use the same name then it skips the point of the test, there will be a second version in the next response. I could use something like `a/bb` to do the same thing without the subdirectory bit.

I'm not sure I see why the subdirectory has any effect though, can you explain more?

ppolewicz: Right. In order to test whether the limit is passed correctly we'd have to mock up the session, I'd say there is no point in doing that. Lets leave it.

ppolewicz:
Ready to merge?

Mossop:
> Ready to merge?

Sure if you're happy with it.

ppolewicz:
Final request - please add an entry to the changelog in `README.md`.

I think we should have a checklist or something so that we don't forget about such things during the earlier phases of review.

Mossop:
Updated, does adding the new function mean needing to bump the minor version?

ppolewicz:
Yes. There were other changes merged in recently (which did not adjust the changelog appropriately) so the next release will be 1.1.0.

I'm going to merge this now. Thank you very much for your contribution.

While we are here, perhaps you would like to contribute some other change (even more challenging, possibly), perhaps more challenging? Or maybe you could help us test the new functionality (4500 LOC PR merged this week, testing it wouldn't hurt) (or code review, for that matter)?

PR #125, opened 2020-05-27
Prevent directory traversal attack in sync functionality

keithele:
The `sync` functionality suffered from a directory traversal attack due to the use of `os.path.join(root_path, file_name)` where the `file_name` could contain a relative path or absolute path outside the `root_path` specified.  This would allow an attacker to write files to arbitrary locations depending on the contents of the bucket being synchronized.

Solution is two-fold:
 1. Do not allow relative or absolute paths in b2 file names being synchronized
 2. Ensure all files written are within the specified root path

### DIFF
--- a/b2sdk/sync/folder.py
+++ b/b2sdk/sync/folder.py
@@ -10,6 +10,7 @@
 
 import logging
 import os
+import re
 import six
 import sys
 
@@ -124,12 +125,25 @@ class LocalFolder(AbstractFolder):
 
     def make_full_path(self, file_name):
         """
-        Convert a file name into an absolute path.
+        Convert a file name into an absolute path, ensure it is not outside self.root
 
         :param file_name: a file name
         :type file_name: str
         """
-        return os.path.join(self.root, file_name.replace('/', os.path.sep))
+        # Setup the relative_path starting at self.root
+        relative_path = os.path.relpath(file_name.replace('/', os.path.sep), start=self.root)
+
+        # Convert the relative path to an absolute path
+        full_path = os.path.abspath(relative_path)
+
+        # Get the common prefix between the new full_path and self.root
+        common_prefix = os.path.commonprefix([full_path, self.root])
+
+        # Ensure the new full_path is inside the self.root directory
+        if common_prefix != self.root:
+            raise Exception("illegal file name: '" + full_path + "'")
+
+        return full_path
 
     def ensure_present(self):
         """
@@ -301,6 +315,20 @@ class B2Folder(AbstractFolder):
 
             if policies_manager.should_exclude_file(file_name):
                 continue
+
+            # Do not allow relative paths in file names
+            if (file_name.startswith('..' + os.path.sep) or
+                os.path.sep + '..' + os.path.sep in file_name or
+                file_name.endswith(os.path.sep + '..') or file_name == '..'
+            ):
+                raise Exception(
+                    "sync does not support file names that include relative paths: %s" % file_name
+                )
+            # Do not allow absolute paths in file names
+            if file_name.startswith(os.path.sep) or re.match("[A-Za-z]:" + os.path.sep, file_name):
+                raise Exception(
+                    "sync does not support file names with absolute paths: %s" % file_name
+                )
 
             if current_name != file_name and current_name is not None and current_versions:
                 yield File(current_name, current_versions)
--- a/b2sdk/sync/folder.py
+++ b/b2sdk/sync/folder.py
@@ -23,7 +23,7 @@ from .scan_policies import DEFAULT_SCAN_MANAGER
 from ..raw_api import SRC_LAST_MODIFIED_MILLIS
 from ..utils import fix_windows_path_limit, is_file_readable
 
-DRIVE_MATCHER = re.compile(r"^([A-Za-z]):([\\/])")
+DRIVE_MATCHER = re.compile(r"^([A-Za-z]):([\\])")
 ABSOLUTE_PATH_MATCHER = re.compile(r"^(/)|^(\\)")
 RELATIVE_PATH_MATCHER = re.compile(
     r"^(\.\.[/\\])|^(\.[/\\])|([/\\]\.\.[/\\])|([/\\]\.[/\\])|([/\\]\.\.)$|([/\\]\.)$|^(\.\.)$|" +
@@ -334,8 +334,8 @@ class B2Folder(AbstractFolder):
                 raise UnSyncableFilename(
                     "sync does not support file names with absolute paths", file_name
                 )
-            # Do not allow Windows drive letters in file names
-            if DRIVE_MATCHER.search(file_name):
+            # On Windows, do not allow drive letters in file names
+            if platform.system() == "Windows" and DRIVE_MATCHER.search(file_name):
                 raise UnSyncableFilename(
                     "sync does not support file names with drive letters", file_name
                 )
--- a/b2sdk/sync/folder.py
+++ b/b2sdk/sync/folder.py
@@ -23,7 +23,7 @@ from .scan_policies import DEFAULT_SCAN_MANAGER
 from ..raw_api import SRC_LAST_MODIFIED_MILLIS
 from ..utils import fix_windows_path_limit, is_file_readable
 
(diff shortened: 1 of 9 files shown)


b2sdk/sync/folder.py
```diff
@@ -294,6 +308,20 @@ def all_files(self, reporter, policies_manager=DEFAULT_SCAN_MANAGER):
             if policies_manager.should_exclude_file(file_name):
                 continue
 
+            # Do not allow relative paths in file names
+            if (file_name.startswith('..' + os.path.sep) or
```
ppolewicz: I think a better way to implement it would be to pass it into a regular expression containing a few `|`

b2sdk/sync/folder.py
```diff
@@ -294,6 +308,20 @@ def all_files(self, reporter, policies_manager=DEFAULT_SCAN_MANAGER):
...
+                os.path.sep + '..' + os.path.sep in file_name or
+                file_name.endswith(os.path.sep + '..') or file_name == '..'
+            ):
+                raise Exception(
+                    "sync does not support file names that include relative paths: %s" % file_name
+                )
+            # Do not allow absolute paths in file names
+            if file_name.startswith(os.path.sep) or re.match("[A-Za-z]:" + os.path.sep, file_name):
```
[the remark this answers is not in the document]
keithele: Moving it to the top.  Luckily most Python implementations cache the compiled regular expressions.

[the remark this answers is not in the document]
keithele: There will always be issues stemming from the differences in the namespaces between the  b2 namespace, Windows filesystem namespace, and linux or OSX's filesystem namespace.  Since b2's namespace is, in many ways, more permissive than either Windows, OSX, or LInux, potential issues can arise.

It is up to the users of the b2 sync function to ensure they do not violate the namespace limitations for the the platform namespaces they intend to use.  It's easy to upload a file with a colon (':') in it and break syncing with Windows.  It's even *easier* to go from a case-sensitive namespace to a non-case-sensitive namespace and cause problems.

In this case, we are not trying to solve/avoid/limit namespace issues, we simply wish to prevent the sync function from writing files outside the designated local destination directory.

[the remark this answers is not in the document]
ppolewicz: ok, fair point, but then `c:/abc` is not an absolute path on linux and this line is refusing to sync it. I think we need to check the platform and base the allowed paths on the result of that check.

b2sdk/sync/folder.py
```diff
@@ -123,12 +124,25 @@ def all_files(self, reporter, policies_manager=DEFAULT_SCAN_MANAGER):
...
+        relative_path = os.path.relpath(file_name.replace('/', os.path.sep), start=self.root)
+
+        # Convert the relative path to an absolute path
+        full_path = os.path.abspath(relative_path)
+
+        # Get the common prefix between the new full_path and self.root
+        common_prefix = os.path.commonprefix([full_path, self.root])
+
+        # Ensure the new full_path is inside the self.root directory
+        if common_prefix != self.root:
```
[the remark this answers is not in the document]
keithele: I've updated it with test cases which I believe cover these cases.

b2sdk/sync/folder.py
```diff
@@ -123,12 +124,25 @@ def all_files(self, reporter, policies_manager=DEFAULT_SCAN_MANAGER):
...
+            raise Exception("illegal file name: '" + full_path + "'")
```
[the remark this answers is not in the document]
keithele: Agreed. I was following the example of this similar type of error raising Exception in the existing code at folder.py:156, 158, & 207.

Updated with UnSyncableFilename error class.

b2sdk/sync/folder.py
keithele: Updated as suggested.

b2sdk/sync/folder.py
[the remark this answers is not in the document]
keithele: Agreed.  The driver-letter matching is now only applied on Windows.

ppolewicz:
@keithele please run `./pre-commit.sh` again so that yapf can unify the formatting again

keithele:
pre-commit.sh run & committed.

b2sdk/sync/folder.py
```diff
@@ -10,17 +10,26 @@
...
 from .scan_policies import DEFAULT_SCAN_MANAGER
 from ..raw_api import SRC_LAST_MODIFIED_MILLIS
 from ..utils import fix_windows_path_limit, is_file_readable
 
+DRIVE_MATCHER = re.compile(r"^([A-Za-z]):([/\\])")
+ABSOLUTE_PATH_MATCHER = re.compile(r"^(/)|^(\\)")
+RELATIVE_PATH_MATCHER = re.compile(
+    r"^(\.\.[/\\])|^(\.[/\\])|([/\\]\.\.[/\\])|([/\\]\.[/\\])|([/\\]\.\.)$|([/\\]\.)$|^(\.\.)$|" +
+    r"([/\\][/\\])|^(\.)$"
+)
```
ppolewicz, line 31:
```suggestion
RELATIVE_PATH_MATCHER = re.compile(
                           # "abc" and "xyz" represent anything, including "nothing"
    r"^(\.\.[/\\])|" +     # ../abc or ..\abc
    r"^(\.[/\\])|" +       # ./abc or .\abc
    r"([/\\]\.\.[/\\])|" + # abc/../xyz or abc\..\xyz or abc\../xyz or abc/..\xyz
    r"([/\\]\.[/\\])|" +   # abc/./xyz or abc\.\xyz or abc\./xyz or abc/.\xyz
    r"([/\\]\.\.)$|" +     # abc/.. or abc\..
    r"([/\\]\.)$|" +       # abc/. or abc\. 
    r"^(\.\.)$|" +         # just ".."
    r"([/\\][/\\])|" +     # abc\/xyz or abc/\xyz or abc//xyz or abc\\xyz
    r"^(\.)$"              # just "."
)  # yapf: disable
```
I have translated the regex into somewhat more readable patterns and now that I look at it, it seems that it will not work correctly in some cases. For example, `abc\.\xyz` is a perfectly legal file name on linux, so a file with such name synced well so far and would have crashed the sync after this change is merged. I don't think crashing it would be good in such case.

To fix it, I think we need to a separate regular expression for Windows and non-windows systems.

The absolute matcher is also affected.

test/sync/test_base.py
```diff
@@ -0,0 +1,36 @@
...
+# License https://www.backblaze.com/using_b2_code.html
+#
+######################################################################
+
+from contextlib import contextmanager
+import re
+import unittest
+
+
+class TestBase(unittest.TestCase):
```
ppolewicz, line 16: isn't this file a copy of another file? Could it not have been imported?

keithele: When working on a new project, I follow the existing conventions.  In this case, I followed the existing convention of the other three existing `test_base.py` files.

I think we're veering off the point of this PR, which was to prevent the directory travesal attack, not refactor all the existing code.

b2sdk/sync/folder.py
keithele: Separate regular expressions will lead to further security vulnerabilities.  Using separate OS specific filters assumes the b2 file names only contain the local OS path separators.  Path separator conversion is performed in other places throughout the code (which could change any time in the future).  By dividing the regular expressions into platform-specific lists, malicious paths in a different platform's format will bypass the local-specific check and then be converted into a locally-malicious path later in the code.  Keeping the expressions together covers all the cases in one place.

Concerning the use of the "." path, the design is to disallow relative paths.  "." is a relative path and is not needed, nor desired, and would lead to sync problems.  Consider this list of perfectly legal files in b2:

"a/b/file.txt"
"a/b/./file.txt"

This example would not sync properly and should be disallowed.

ppolewicz: I just checked and the old sync could not have synced those names due to the conversions you've mentioned. 👍

test/sync/test_base.py
ppolewicz: I'll file a separate issue for it

ppolewicz: https://github.com/Backblaze/b2-sdk-python/issues/131

PR #128, opened 2020-05-28
Sync from timestamp

mlech-reef:


b2sdk/sync/folder.py
```diff
@@ -230,11 +232,16 @@ def _walk_relative_paths(self, local_dir, b2_dir, reporter, policies_manager):
                 # to iterate through large folders
                 if is_file_readable(local_path, reporter):
                     file_mod_time = int(os.path.getmtime(local_path) * 1000)
+
+                    if policies_manager.should_exclude_file_version(file_mod_time):
```
ppolewicz, line 236: I think a policy manager may decide to exclude a file version based on something that is not `file_mod_time`. We should either rename that method so that it is obvious it filters mod times, or, preferably, make it accept a FileVersion object which will allow for more sophisticated logic, using size, path, mod time etc.

b2sdk/sync/folder.py
```diff
@@ -294,7 +301,7 @@ def all_files(self, reporter, policies_manager=DEFAULT_SCAN_MANAGER):
             if policies_manager.should_exclude_file(file_name):
                 continue
 
-            if current_name != file_name and current_name is not None:
+            if current_name != file_name and current_name is not None and current_versions:
```
ppolewicz, line 304: that's worth unit testing

b2sdk/sync/folder.py
```diff
@@ -303,13 +310,19 @@ def all_files(self, reporter, policies_manager=DEFAULT_SCAN_MANAGER):
             else:
                 mod_time_millis = file_version_info.upload_timestamp
             assert file_version_info.size is not None
+
+            current_name = file_name
+
+            if policies_manager.should_exclude_file_version(mod_time_millis):
```
ppolewicz, line 316: same as before, this method should accept an object, not an integer

b2sdk/sync/folder.py
mlech-reef: I totally agree, but to be consequent, your comment can be also applied to `should_exclude_file` and `should_exclude_directory`. They should have different names or, preferably, accept `File` and `AbstractFolder`. In the beginning, I wanted to do it that way, but the folder object is only created for the root and we would have to rewrite the logic completely, in NBC manner. What's more, for a large directory structure, it could cause slowing down, by creating folder objects just to pass it to the `ScanPoliciesManager`.

I can change the signature of the `should_exclude_file_version`, but it won't be aligned with the rest of `ScanPoliciesManager`. Or I can change the name. You call. For me, the current API is ok, aligned with the rest.

b2sdk/sync/folder.py
mlech-reef: test_exclude_modified_multiple_versions

mlech-reef: I added one more to better cover if from line 325

b2sdk/sync/folder.py
ppolewicz: this was discussed separately and NBC interface has won

PR #129, opened 2020-05-30
Add support for XDG_CONFIG_HOME

0az:
Implement support for `XDG_CONFIG_HOME`.

To avoid breaking changes, we use locations in this order:
- `$B2_ACCOUNT_INFO_ENV_VAR`: If set
- `~/.b2_account_info`: If exists
- `$XDG_CONFIG_HOME/b2/account_info`: If set
- `~/.b2_account_info`: If no path takes priority


ppolewicz:
We used to be able to pass a file path to the constructor of SqliteAccountInfo, but you removed it. Please restore it.

Also, could you briefly explain why are you proposing this change?

0az:
> We used to be able to pass a file path to the constructor of SqliteAccountInfo, but you removed it. Please restore it.

Passing in the file path still works: I just forgot to add it to the explanatory text and the docstring. I have since done the latter.

> Also, could you briefly explain why are you proposing this change

I'm proposing the change so I can remove another dotfile from `$HOME` and add b2sdk/b2 cli to the list of applications that support the XDG spec. If I'm not mistaken, version 1.1.0 should go out soonish, and will be used in the next version of the b2 cli.

ppolewicz:
I think it is a good change. Could you please add a unit test to make sure that we don't hit a regression on this new functionality in the future?

0az:
I added unit tests for two cases:
- `XDG_CONFIG_HOME`
- `B2_ACCOUNT_INFO_ENV_VAR` and `XDG_CONFIG_HOME` are set.

I'm not sure how to test the third interaction, where `XDG_CONFIG_HOME` is set, but a database already exists. This may require changing the `file_name` kwarg to override `B2_ACCOUNT_INFO_DEFAULT_FILE` instead of overriding everything.

A hermetic testing environment and compatibility with existing code is ensured by modifying `_make_sqlite_account_info` to take an `env` kwarg. If it is unset or falsy, it will instantiate SqliteAccountInfo with an empty env. If it is truthy, it will override the usual `file_name` behavior.

b2sdk/account_info/sqlite_account_info.py
```diff
@@ -34,17 +34,28 @@ class SqliteAccountInfo(UrlPoolAccountInfo):
...
 
         :param str file_name: The sqlite file to use; overrides the default.
+        :param dict env: Override Environment variables. For testing only.
         :param int last_upgrade_to_run: For testing only, override the auto-update on the db.
         """
         self.thread_local = threading.local()
-        user_account_info_path = file_name or os.environ.get(
-            B2_ACCOUNT_INFO_ENV_VAR, B2_ACCOUNT_INFO_DEFAULT_FILE
-        )
+        if B2_ACCOUNT_INFO_ENV_VAR in env:
```
bwbeach, line 50: I think this code works, but the logic would be clearer if it mirrored the structure of the comment, above:

```
if file_name:
    use it
elif B2_ACCOUNT_INFO_ENV_VAR in env:
    use it
elif default file exists:
    use it
elif XDG_CONFIG_HOME in env:
    use it
else:
    use default path
```

test/v1/test_account_info.py
```diff
@@ -292,11 +292,39 @@ def test_upgrade_2_default_app_key(self):
...
+    def _make_sqlite_account_info(self, env=None, last_upgrade_to_run=None):
         """
-        Returns a new StoredAccountInfo that has just read the data from the file.
+        Returns a new SqliteAccountInfo that has just read the data from the file.
         """
-        return SqliteAccountInfo(file_name=self.db_path, last_upgrade_to_run=None)
+        return SqliteAccountInfo(
+            file_name=self.db_path if not env else None,
+            env=env or {},
+            last_upgrade_to_run=last_upgrade_to_run
```
ppolewicz, line 302: please add a trailing `,`  in this line so that `yapf` reliably guards the format of the multi-line statement

test/v1/test_account_info.py
```diff
@@ -292,11 +292,39 @@ def test_upgrade_2_default_app_key(self):
...
+        )
 
     def test_account_info_persistence(self):
         self._test_account_info(check_persistence=True)
+
+    def test_uses_xdg_config_home(self):
+        with tempfile.TemporaryDirectory() as d:
```
ppolewicz, line 309: `tempfile.TemporaryDirectory` was added in 3.2, but we still support 2.7, but we do have [b2sdk.utils.TempDir](https://github.com/Backblaze/b2-sdk-python/blob/43b42e7562ec82c8baefd2614b1be6aefbfb0e0e/b2sdk/utils.py#L239) context manager. Please switch to it.

test/v1/test_account_info.py
```diff
@@ -292,11 +292,39 @@ def test_upgrade_2_default_app_key(self):
...
+            account_info = self._make_sqlite_account_info(env={'XDG_CONFIG_HOME': d})
+            expected_path = os.path.abspath(os.path.join(d, 'b2', 'account_info'))
+            actual_path = os.path.abspath(account_info.filename)
+            assert expected_path == actual_path, 'Actual path %s is not equal to $XDG_CONFIG_HOME/b2/account_info'
```
ppolewicz, line 313:
```suggestion
            assert expected_path == actual_path, 'Actual path %s is not equal to $XDG_CONFIG_HOME/b2/account_info' % (actual_path,)
```

test/v1/test_account_info.py
```diff
@@ -292,11 +292,39 @@ def test_upgrade_2_default_app_key(self):
...
+                env={
+                    'XDG_CONFIG_HOME': d,
+                    B2_ACCOUNT_INFO_ENV_VAR: os.path.join(d, 'b2_account_info')
+                }
+            )
+            expected_path = os.path.abspath(os.path.join(d, 'b2_account_info'))
+            actual_path = os.path.abspath(account_info.filename)
+            assert expected_path == actual_path, 'Actual path {path} is not equal to ${var}'.format(
+                path=actual_path, var=B2_ACCOUNT_INFO_ENV_VAR
+            )
```
ppolewicz, line 330:
```suggestion
            assert expected_path == actual_path, 'Actual path %s is not equal to %s' % (
                actual_path,
                B2_ACCOUNT_INFO_ENV_VAR,
            )
```
Most of the code uses % for string formatting, I've only now discovered that one new module uses `.format()` and I have filed an issue to fix it. We should keep to the `%`, at least until we drop 3.5, so that the code is consistent across the project.

test/v1/test_account_info.py
```diff
@@ -292,11 +293,45 @@ def test_upgrade_2_default_app_key(self):
...
 
     def test_account_info_persistence(self):
         self._test_account_info(check_persistence=True)
+
+    def test_uses_xdg_config_home(self):
+        with TempDir() as d:
+            account_info = self._make_sqlite_account_info(env={'XDG_CONFIG_HOME': d})
+            expected_path = os.path.abspath(os.path.join(d, 'b2', 'account_info'))
+            actual_path = os.path.abspath(account_info.filename)
+            assert expected_path == actual_path, 'Actual path %s is not equal to $XDG_CONFIG_HOME/b2/account_info' % (
```
ppolewicz, line 317: I think you can use `self.assertEquals` here, it's a bit cleaner

ppolewicz: in the next test too

b2sdk/account_info/sqlite_account_info.py
```diff
@@ -34,19 +34,41 @@ class SqliteAccountInfo(UrlPoolAccountInfo):
...
+            user_account_info_path = file_name
+        elif B2_ACCOUNT_INFO_ENV_VAR in os.environ:
+            user_account_info_path = os.environ[B2_ACCOUNT_INFO_ENV_VAR]
+        elif os.path.exists(os.path.expanduser(B2_ACCOUNT_INFO_DEFAULT_FILE)):
+            user_account_info_path = B2_ACCOUNT_INFO_DEFAULT_FILE
+        elif 'XDG_CONFIG_HOME' in os.environ:
+            config_home = os.environ['XDG_CONFIG_HOME']
+            user_account_info_path = os.path.join(config_home, 'b2', 'account_info')
+            if not os.path.exists(os.path.join(config_home, 'b2')):
+                os.makedirs(os.path.join(config_home, 'b2'), mode=0o755, exist_ok=True)
```
ppolewicz, line 65: this was actually not resolved, will fail on 2.7

test/v1/test_account_info.py
```diff
@@ -292,11 +293,48 @@ def test_upgrade_2_default_app_key(self):
...
-        Returns a new StoredAccountInfo that has just read the data from the file.
+        Returns a new SqliteAccountInfo that has just read the data from the file.
+
+        :param dict env: Override Environment variables.
         """
-        return SqliteAccountInfo(file_name=self.db_path, last_upgrade_to_run=None)
+        with mock.patch.dict('os.environ', env or {}):
+            return SqliteAccountInfo(
+                file_name=self.db_path if not env else None,
+                env=env or {},
```
ppolewicz, line 305: you have removed the `env` argument from `SqliteAccountInfo` constructor, but it remained here in tests, causing all tests of `SqliteAccountInfo` to fail. Could you please fix it and run `./pre-commit.sh` to make sure all tests pass before we merge it in?

Also, could you please add `Add support for XDG_CONFIG_HOME (0az)` to changelog section in readme?

0az:
I'm really sorry about that.

I probably should have mentioned this earlier, but for some unknown reason that I don't have time to debug after midnight, the pre-commit shell script fails with a RuntimeError: "An attempt has been made to start a new process before the current process has finished its bootstrapping phase." Logs are in this gist: https://gist.github.com/0az/234a94a123e15259f3779661feed69e6. Manually running nosetests still works as a workaround, though.

I also modified the test class to include a temporary directory acting as HOME.

ppolewicz:
@mlech-reef did we accidentally enable parallel testing in nosetests on Windows? I don't think that ever worked, now appveyor can't run the tests and our contributor here also has issues. Travis is absent from this PR for some reason... Do you have any clue?

mlech-reef:
Travis is triggered properly, but for some reason, it's not showing here. I've seen this issue in other PRs. We should investigate it.

mlech-reef:
> @mlech-reef did we accidentally enable parallel testing in nosetests on Windows? I don't think that ever worked, now appveyor can't run the tests and our contributor here also has issues.

I don't recall anything like that. AppVeyor fails not because of the parallel runs but because 2 new tests are failing with:
`PermissionError: [WinError 32] The process cannot access the file because it is being used by another process`
It's probably because `TempDir` uses `shutils.rmtree` on a directory with sqlite db file created inside.
It's similar to the problem described here: https://stackoverflow.com/questions/1213706/what-user-do-python-scripts-run-as-in-windows
Windows and python sometimes don't get along :)

b2sdk/account_info/sqlite_account_info.py
```diff
@@ -36,17 +36,39 @@ class SqliteAccountInfo(UrlPoolAccountInfo):
 
     def __init__(self, file_name=None, last_upgrade_to_run=None):
         """
-        If ``file_name`` argument is empty or ``None``, path from ``B2_ACCOUNT_INFO`` environment variable is used. If that is not available, a default of ``~/.b2_account_info`` is used.
```
ppolewicz, line 39: I am reading this docstring over and over and I think apiver doesn't let us change the default location to use `XDG_CONFIG_HOME` in v1. Someone may have written a program that relies on the promise that when the filename was not passed and `B2_ACCOUNT_INFO` was not set, then the db file will land in `~/.b2_account_info`. After running something, the program might clean the environment by removing that file.

If we make this change and `XDG_CONFIG_HOME` is defined, sdk user would not expect the configuration of the end user to be suddenly leaked (as it is not cleaned up anymore), possibly allowing access to the following user.

I think this is a good change and we should introduce it, but that should happen with the release of `from b2sdk.v2 import *`, otherwise we risk introducing a breaking change, while the purpose of apiver is to avoid it unless fixing a bug (which we are not doing here) which would change behavior because the old behavior was just wrong.

Furthermore, we should separate the documentation of the interface from the documentation of the current behavior, so that this can be altered in the future, by mentioning that the user should not define of this algorithm staying exactly the same, but should use `self.filename` to determine the actually determined location of the file in case it needs to be accessed later.

We did not have a compelling reason to release v2, but this is a good one.

@bwbeach @0az do you agree with my viewpoint? Shall we move it to v2?

0az: Good catch: I didn't actually think of that case.

Sounds good to me.

ppolewicz: @0az then would you please adjust the docstring here? We'll hold off merging this until b2sdk 1.1.0 and b2cli 2.0.0 are shipped, then we'll increment apiver to v2 and finally we'll merge your thing (though a final adjustment will be necessary to provide backwards compatibility)

0az: Another question for v2: should SqliteAccountInfo support `:memory:` or anonymous (empty string) databases?

https://www.sqlite.org/c3ref/open.html
> If the filename is ":memory:", then a private, temporary in-memory database is created for the connection. This in-memory database will vanish when the database connection is closed. Future versions of SQLite might make use of additional special filenames that begin with the ":" character. It is recommended that when a database filename actually does begin with a ":" character you should prefix the filename with a pathname such as "./" to avoid ambiguity.
> 
> If the filename is an empty string, then a private, temporary on-disk database will be created. This private database will be automatically deleted as soon as the database connection is closed.

b2sdk/account_info/sqlite_account_info.py
```diff
@@ -36,17 +36,39 @@ class SqliteAccountInfo(UrlPoolAccountInfo):
...
+        Initialize SqliteAccountInfo.
+
+        Locations are used in the following order:
```
0az, line 41:
```suggestion
        The exact algorithm used to determine the location of the database file is not API in any sense.
        If the location of the database file is required (for cleanup, etc), do not assume a specific resolution:
        instead, use ``self.filename`` to get the actual resolved location.

        SqliteAccountInfo currently checks locations in the following order:
```
@ppolewicz How does this look?

b2sdk/account_info/sqlite_account_info.py
ppolewicz: It looks like that could replace `v` with one line of code. I like less code.

b2sdk/account_info/sqlite_account_info.py
ppolewicz: that's way better, thanks

mpnowacki-reef:
@0az I have submitted a PR that merges master into your `xdg` branch and moves your changes along with tests to v2: https://github.com/0az/b2-sdk-python/pull/1/
We're planning to wrap up v2 by the end of this week, would you be willing to finish this PR?

0az:
Thanks! Was there anything else that I needed to do? It's been a while. 🙃

b2sdk/v1/account_info.py
```diff
@@ -76,7 +79,24 @@ class UrlPoolAccountInfo(v2.UrlPoolAccountInfo, AbstractAccountInfo):
...
+        :param str file_name: The sqlite file to use; overrides the default.
+        :param int last_upgrade_to_run: For testing only, override the auto-update on the db.
+        """
+        # use legacy env var resolution, XDG not supported
+        self.thread_local = threading.local()
+        user_account_info_path = file_name or os.environ.get(
+            v2.B2_ACCOUNT_INFO_ENV_VAR, v2.B2_ACCOUNT_INFO_DEFAULT_FILE
+        )
+        self.filename = file_name or os.path.expanduser(user_account_info_path)
+        logger.debug('%s file path to use: %s', self.__class__.__name__, self.filename)
```
ppolewicz, line 95: where is `logger` initialized?

ppolewicz (changes requested):

ppolewicz:
This PR seems to contain v2 changes. Can we somehow make it clean? @0az would you mind if we took your changes and rebased them to slide it into v2 at the appropriate moment? The v2 change is a little bit chaotic and I don't want to keep pulling you into this.

0az:
> This PR seems to contain v2 changes. Can we somehow make it clean?

Michał sent in a PR with the v2 changes, so I merged that in. I can force push that off of the branch, if you want.

> @0az would you mind if we took your changes and rebased them to slide it into v2 at the appropriate moment? The v2 change is a little bit chaotic and I don't want to keep pulling you into this.

This is also fine.

b2sdk/v1/account_info.py
0az: It's imported from here:

https://github.com/Backblaze/b2-sdk-python/blob/632b13b74aa584a9c962a9b85f4630b8373cba1d/b2sdk/account_info/sqlite_account_info.py#L23

ppolewicz: ok, but here in `b2sdk/v1/account_info.py` there is no `logger`, so it will crash. Logger initialization should be added.

0az: Sorry, bad wording. The initialization happens there, and there's an import in this file *for* that logger.

It probably should get initialized here anyway, though.

ppolewicz: `from b2sdk.account_info.sqlite_account_info import logger` ohhh. I don't think we should do that, that is super confusing. We should initialize it locally, right?

mpnowacki-reef:
@0az I've come to clean up the mess I made. I squashed all your commits on top of the current master and created a new PR (https://github.com/Backblaze/b2-sdk-python/pull/235/). I have made you the author of the first commit, in order not to steal your thunder. Is that okay with you?

0az:
Yeah, 235 looks good. I'll close this PR, then.

PR #138, opened 2020-06-19
Fix to use SHA1 of upload source if it is known

mzukowski-reef:


ppolewicz:
https://github.com/Backblaze/B2_Command_Line_Tool/issues/539 is related

PR #140, opened 2020-06-29
Fix upload part or file range on Python 2.7

mzukowski-reef:
Python 2.7 `open()` is returning object with
`seek()` implementation returning `None`.

Fixed it with using `io.open()` for file upload source.
Added simulation of chunked read with retry in RawSimulator for uploading,
because all streams have to support seeking to beginning after reading
entire stream.

Resolves Backblaze/B2_Command_Line_Tool#639


ppolewicz:
We have a solution, but it messed up the tests, so we need to fix them too to release the fix

PR #141, opened 2020-06-30
Fix WritingStreamWithProgress

mzukowski-reef:
Remove `seek()` progress updating logic - writes can happen
in different parts of file.

Fix tests to reliably check for non-monotonic download progress

### DIFF
--- a/b2sdk/raw_simulator.py
+++ b/b2sdk/raw_simulator.py
@@ -307,7 +307,9 @@ class FakeResponse(object):
 
     def iter_content(self, chunk_size=1):
         start = 0
+        rnd = random.Random(self.url)
         while start <= len(self.data_bytes):
+            time.sleep(rnd.random() * 0.01)
             yield self.data_bytes[start:start + chunk_size]
             start += chunk_size
 
--- a/b2sdk/stream/progress.py
+++ b/b2sdk/stream/progress.py
@@ -60,6 +60,12 @@ class ReadingStreamWithProgress(AbstractStreamWithProgress):
 
     def seek(self, pos, whence=0):
         pos = super(ReadingStreamWithProgress, self).seek(pos, whence=whence)
+        # reset progress to current stream position - assumption is that ReadingStreamWithProgress would not be used
+        # for random access streams, and seek is only used to reset stream to beginning to retry file upload
+        # and multipart file upload would open and use different file descriptor for each part;
+        # this logic cannot be used for WritingStreamWithProgress because multipart download has to use
+        # single file descriptor and synchronize writes so seeking cannot be understood there as progress reset
+        # and writing  progress is always monotonic
         self.bytes_completed = pos
         return pos
 
--- a/b2sdk/stream/progress.py
+++ b/b2sdk/stream/progress.py
@@ -32,11 +32,6 @@ class AbstractStreamWithProgress(StreamWrapper):
         self.bytes_completed = 0
         self.offset = offset
 
-    def seek(self, pos, whence=0):
-        pos = super(AbstractStreamWithProgress, self).seek(pos, whence=whence)
-        self.bytes_completed = pos
-        return pos
-
     def _progress_update(self, delta):
         self.bytes_completed += delta
         self.progress_listener.bytes_completed(self.bytes_completed + self.offset)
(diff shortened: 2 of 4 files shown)


test/v1/test_bucket.py
```diff
@@ -714,20 +730,28 @@ def _make_data(self, approximate_length):
 
 # Downloads
 
-
-class DownloadTests(object):
-    DATA = 'abcdefghijklmnopqrs'
+class DownloadTestsBase(object):
+    DATA = ''
```
ppolewicz, line 734:
```suggestion
    DATA = NotImplemented
```

b2sdk/stream/progress.py
```diff
@@ -63,6 +58,11 @@ def read(self, size=None):
         self._progress_update(len(data))
         return data
 
+    def seek(self, pos, whence=0):
+        pos = super(AbstractStreamWithProgress, self).seek(pos, whence=whence)
```
ppolewicz, line 62:
```suggestion
        pos = super(ReadingStreamWithProgress, self).seek(pos, whence=whence)
```

b2sdk/raw_simulator.py
```diff
@@ -307,7 +307,9 @@ def __init__(self, file_sim, url, range_=None):
 
     def iter_content(self, chunk_size=1):
         start = 0
+        rnd = random.Random(self.url)
```
bwbeach, line 310: Why seed with the url?  So it's reproducible?

b2sdk/stream/progress.py
```diff
@@ -63,6 +58,11 @@ def read(self, size=None):
         self._progress_update(len(data))
         return data
 
+    def seek(self, pos, whence=0):
+        pos = super(ReadingStreamWithProgress, self).seek(pos, whence=whence)
+        self.bytes_completed = pos
```
bwbeach, line 63: Given that this was the source of a bug, it would be nice to have a comment here saying why 
it's appropriate to set `bytes_completed` here, but not in `AbstractStreamWithProgress`.

bwbeach:
I made a couple comments on the changes.  

It's not clear to me what the bug was.  Did it have to do with writes happening out of order on the file?  If so, how is the progress updating supposed to work?  Is it worth having some unit tests for `ReadingStreamWithProgress` and `WritingStreamWithProgress`?

b2sdk/raw_simulator.py
mzukowski-reef: Yes - the idea is that if I run test case two times it should run exactly the same way... we can either hardcode some arbitrary value here or use something from test case context... like file download url - I like more the second option :) but actually this is for forcing nonsequential writes in parallel download... and thread context switching is not deterministic... so it actually probably doesn't matter so much - but it is my pattern to seed random number generator from test case context...

b2sdk/stream/progress.py
mzukowski-reef: added comment

mzukowski-reef:
> It's not clear to me what the bug was. Did it have to do with writes happening out of order on the file? If so, how is the progress updating supposed to work? Is it worth having some unit tests for `ReadingStreamWithProgress` and `WritingStreamWithProgress`?

@bwbeach - the bug was that progress cannot be reset on seek in `WritingStreamWithProgress` because in multipart download single file descriptor has to be used and writes are synchronized in single thread, and before each write seek is used - so progress can only be updated on `write` and is always monotonic - if for some reason application would write multiple times to the same file position, then write progress would exceed file size - it was a case with `ReadingStreamWithProgress` - we have `seek(0)` before starting file upload and upload can be retried multiple times in b2http (it can raise error up and then file would be reopened, but some errors can cause retry on b2http level) - when fixing a bug with seek on Python 2.7 (#140) - I've decided to simulate b2http retry in `RawSimulator` - so we would check in every test case if stream used in test case supports b2http retry. It caused some errors with `ReadingStreamWithProgress` - and this was actually a bug, that in case of retry progress listener was showing progress outside upload content length - now `ReadingStreamWithProgress` is reporting correct progress and progress listener can either rewind progress or wait until progress pass value before reset (that is how `TqdmProgressListener` is working) - but as I said I've made a mistake and added change to `AbstractStreamWithProgress` - which caused issue in `WritingStreamWithProgress`

There are tests for progress listener in bucket tests - after those two changes progress is better tested (both for uploads and downloads) - so I think that there is no need for more dedicated unit tests.

ppolewicz:
The fix was placed too high in inheritance chain and affected a class that it should not have affected.

It may be possible to one day implement a version of WritingStreamWithProgress which would properly understand re-writes (it would have to know the part sizes and track a marker in each part), but the current callers make sure to not write twice to the same byte by only downloading the part of the part that happened after the download was interrupted, so WritingStreamWithProgress that properly understand re-writes would be practically useless. It's better to optimize it on the download side to conserve network resources, than on the writer to optimize progress reporting.

PR #146, opened 2020-07-21
Drop Python 2 and Python 3.4 support

mlech-reef:


README.md
```diff
@@ -39,6 +33,8 @@ b2sdk>=0.0.0,<1.0.0
 
 ## Not released yet
 
+* Drop Python 2 and Python 3.4 support :tada:
```
ppolewicz, line 36: please add a link to https://www.backblaze.com/b2/docs/quick_command_line.html

b2sdk/raw_simulator.py
```diff
@@ -1038,8 +1033,7 @@ def list_buckets(
 
         # Do the query
         sorted_buckets = [
-            self.bucket_name_to_bucket[name]
-            for name in sorted(six.iterkeys(self.bucket_name_to_bucket))
+            self.bucket_name_to_bucket[name] for name in sorted(self.bucket_name_to_bucket)
```
ppolewicz, line 1036: wouldn't sorting output of `.items()` be faster here due to not needing to first go through N to get every key and then retrieving every value with the theoretical "O(1)" which really is O(log N)? I mean, there's a max of 100 entries in this thing anyway so maybe it doesn't matter much, but I'm wondering why you chose to write it this way

b2sdk/sync/folder.py
```diff
@@ -282,12 +280,8 @@ def _handle_non_unicode_file_name(cls, name):
         names.
         """
         # if it's all ascii, allow it
-        if six.PY2:
-            if all(ord(c) <= 127 for c in name):
-                return name
-        else:
-            if all(b <= 127 for b in name):
-                return name
+        if all(b <= 127 for b in name):
```
ppolewicz, line 283: by the way, isn't there a faster way to do this using `str.encode('ascii')` and catching an exception?

b2sdk/raw_simulator.py
mlech-reef: I didn't choose to write it this way :) For most parts, I used regexes and structural replacement to get rid of six. Then I did a review for myself without analyzing the code for quality and speed, only for correctness.

I will update it

mlech-reef: It did what six is doing under the hood for Python 3 with few exceptions, so it didn't introduce any regression.

b2sdk/sync/folder.py
mlech-reef: Trying to decode bytes is a lot slower. This method should not be called if `self.root` is `str`. In my opinion, we should prevent using bytes for paths and raise an exception in such case

b2sdk/raw_simulator.py
mlech-reef: Using `items()` is actually 2x slower because of tuple lookups and unpacking

ppolewicz: 🤦

b2sdk/sync/folder.py
ppolewicz: yeah I think we should move it all to pathlib in v2 and leave wrappers for strings for v1. I updated the ticket with the list of things to consider for apiver v2.

README.md
ppolewicz: actually, I was wrong, that link is for b2cli, not for b2sdk!

mlech-reef: I have removed the last commit

PR #149, opened 2020-08-03
Fixed small typo

ZuluPro:
There was a confusion between `file_info`/`file_infos` in methods' args and docstring


ppolewicz:
Thanks for the patch, feel free to do more of those, the documentation always needs some love!

PR #153, opened 2020-09-03
Add support for unverified sha1

mlech-reef:
Fixes Backblaze/B2_Command_Line_Tool#657


b2sdk/transfer/inbound/file_metadata.py
```diff
@@ -59,3 +59,9 @@ def as_info_dict(self):
             'contentSha1': self.content_sha1,
             'fileInfo': self.file_info,
         }
+
+    @classmethod
+    def _parse_content_sha1(cls, content_sha1):
```
ppolewicz, line 64: I think this method should return a tuple like `(sha1, verified)` and the caller might sha1, _ = cls._parse_content_sha1(content_sha1)` if they don't care, but one they a caller might care and then we should not need to rewrite the parser

b2sdk/transfer/inbound/file_metadata.py
```diff
@@ -59,3 +59,9 @@ def as_info_dict(self):
...
+        if content_sha1.startswith('unverified:'):
+            return content_sha1.replace('unverified:', '')
```
ppolewicz, line 66:
```suggestion
        UNVERIFIED_CHECKSUM_PREFIX = 'unverified:'  # up in class constants or module
        if content_sha1.startswith(cls.UNVERIFIED_CHECKSUM_PREFIX):
            return content_sha1[len(cls.UNVERIFIED_CHECKSUM_PREFIX):]
```

b2sdk/transfer/inbound/file_metadata.py
```diff
@@ -35,7 +35,7 @@ def __init__(
         self.file_name = file_name
         self.content_type = content_type
         self.content_length = content_length
-        self.content_sha1 = content_sha1
+        self.content_sha1 = self._parse_content_sha1(content_sha1)
```
ppolewicz, line 38: should FileMetadata get a new field that tells whether the checksum we are dealing with here is verified or not? Seems easy to add, it's small but can be important for some callers.

test/v0/test_file_metadata.py
```diff
@@ -0,0 +1,35 @@
...
+
+class TestFileMetadata(TestBase):
+    KWARGS = {
+        'file_id': '4_deadbeaf3b3e38a957f100d1e_f1042665d79618ae7_d20200903_m194254_c000_v0001053_t0048',
+        'file_name': 'foo.txt',
+        'content_type': 'text/plain',
+        'content_length': '1',
+        'content_sha1': '4518012e1b365e504001dbc94120624f15b8bbd5',
+        'file_info': {},
+    }
```
ppolewicz, line 28: you need to suppress yapf here

test/v1/test_file_metadata.py
```diff
@@ -0,0 +1,35 @@
...
+
+class TestFileMetadata(TestBase):
+    KWARGS = {
+        'file_id': '4_deadbeaf3b3e38a957f100d1e_f1042665d79618ae7_d20200903_m194254_c000_v0001053_t0048',
+        'file_name': 'foo.txt',
+        'content_type': 'text/plain',
+        'content_length': '1',
+        'content_sha1': '4518012e1b365e504001dbc94120624f15b8bbd5',
+        'file_info': {},
+    }
```
ppolewicz, line 28: you need to suppress yapf here

b2sdk/transfer/inbound/file_metadata.py
mlech-reef: I'm always trying to not have internal code (not part of API) which is not used. It's cleaner, and once we need it we can add it and it's a good chance it won't be needed.

b2sdk/transfer/inbound/file_metadata.py
mlech-reef: Yes, this looks better.

b2sdk/transfer/inbound/file_metadata.py
mlech-reef: Again, it's not part of API so there is no such caller yet.

test/v0/test_file_metadata.py
ppolewicz: I meant it - your formatting was good and yapf broke it. It's better to suppress it than to let it make the code ugly

test/v1/test_file_metadata.py
ppolewicz: here too

b2sdk/transfer/inbound/file_metadata.py
ppolewicz: my point is that the unverified checksum flag should be added as a field of this class, now that we do understand it. Why would we expose the checksum but not it's verification status when the server provides it?

test/v0/test_file_metadata.py
mlech-reef: Whenever we need to do it for such a simple object, it means our yapf configuration is wrong :/

b2sdk/transfer/inbound/file_metadata.py
mlech-reef: Still don't agree, but I can leave with that :)

b2sdk/transfer/inbound/file_metadata.py
mlech-reef: Added

b2sdk/transfer/inbound/file_metadata.py
```diff
@@ -56,6 +59,18 @@ def as_info_dict(self):
...
+    @classmethod
+    def _decode_content_sha1(cls, content_sha1):
+        if content_sha1.startswith(cls.UNVERIFIED_CHECKSUM_PREFIX):
+            return content_sha1[len(cls.UNVERIFIED_CHECKSUM_PREFIX):], False
+        return content_sha1, True
+
+    @classmethod
+    def _encode_content_sha1(cls, content_sha1, content_sha1_verified):
+        if not content_sha1_verified:
+            return '{}{}'.format(cls.UNVERIFIED_CHECKSUM_PREFIX, content_sha1)
```
ppolewicz, line 75: we use `%`-style formatting in this project, could you please change this (for consistency)?

PR #156, opened 2020-09-12
Small clarifications in the docs

eallrich:
While following the tutorial and quick start documentation I noticed a couple possible improvements. Hopefully these contributions are helpful. Thank you for the SDK!

### DIFF
--- a/doc/source/quick_start.rst
+++ b/doc/source/quick_start.rst
@@ -282,6 +282,8 @@ Get file metadata
 Copy file
 =========
 
+Please switch to  :meth:`b2sdk.v1.Bucket.copy`.
+
 .. code-block:: python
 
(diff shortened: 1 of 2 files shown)


doc/source/quick_start.rst
```diff
@@ -282,8 +282,6 @@ Get file metadata
 Copy file
 =========
 
-Please switch to  :meth:`b2sdk.v2.Bucket.copy`.
```
eallrich, line 285: I couldn't import from `b2sdk.v2` and I didn't see any other references to a `v2` of the API so I've pulled this line out. Might this be for an upcoming release instead?

ppolewicz: it should be `b2sdk.v1.Bucket.copy` - v2 is not released yet and we don't need it any time soon

eallrich: Thank you. I've restored the reference and switched it to `v1`.

ppolewicz:
All right, thanks for your contribution, documentation always needs some love!

PR #163, opened 2020-10-07
Fix update bucket

### LINKED ISSUE #162: should bucket.update() let me set my bucket_info to {}?
Hi --

I'd like to set the bucket info on my bucket to not have any entries.  To do that, I tried calling bucket.set_info() and got an assert.  Here's a simple failing case:

> b2_api = B2Api(InMemoryAccountInfo())
> b2_api.authorize_account(url, app_key_id, app_key)
> bucket = b2_api.get_bucket_by_name("myBucket"))
> 
> bucket.set_info({})

This triggers an assert on line 558, in update_bucket:
> assert bucket_info or bucket_type

I'm running on macOS 10.15.7.
I'm using Python 3.8.6 installed with mac ports.
i have the following b2 things installed with pip3:

> $ pip3 list | grep b2
> b2        2.0.2
> b2sdk  1.1.4

Here's the stack trace that's generated: [stack.txt](https://github.com/Backblaze/b2-sdk-python/files/5337715/stack.txt)

A similar thing happens when I use bucket.update(bucket_info={}), which ends up on the same place.  I'm currently working around it by passing the bucket's type back in:
>    bucket.update(bucket_type=bucket.as_dict()['bucketType'], bucket_info={}, lifecycle_rules=[])

mlech-reef:
Fixes #162

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Drop Python 2 and Python 3.4 support :tada:
 * Remove `--prefix` from `ls` (it didn't really work, use `folderName` argument)
 
+### Fixed
+* Allow to set an empty bucket info during the update
+
 ## [1.1.4] - 2020-07-15
 
 ### Added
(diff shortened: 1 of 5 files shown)


ppolewicz (approved):
please add a changelog entry and lets go

PR #165, opened 2020-10-15
Add support for bucket to bucket sync of the latest versions of files

mlech-reef:
An example of usage:
```
$ b2 ls --recursive ML-TestBucket
$ b2 ls --recursive ML-TestBucket2
$ b2 sync ./test b2://ML-TestBucket/test
upload 1.txt
upload 2.txt
upload 3.txt
$ b2 sync b2://ML-TestBucket/test b2://ML-TestBucket2/test
copy 1.txt
copy 2.txt
copy 3.txt
$ b2 ls --recursive ML-TestBucket2
test/1.txt
test/2.txt
test/3.txt
```


b2sdk/bucket.py
```diff
@@ -687,10 +687,10 @@ def copy(
         """
 
         copy_source = CopySource(file_id, offset=offset, length=length)
-        if length is None:
+        if not length:
```
bwbeach, line 690: This check is not the same when `length` is 0.  I think this first branch should still happen in that case.

b2sdk/sync/policy.py
```diff
@@ -305,6 +305,60 @@ class DownAndKeepDaysPolicy(DownPolicy):
...
+        )
+
+
+class CopyAndDeletePolicy(CopyPolicy):
+    """
+    File is copied (server-side) and the delete flag is SET.
+    """
+
+    def _get_hide_delete_actions(self):
+        for action in super()._get_hide_delete_actions():
```
bwbeach, line 332: What actions might the parent class have that we want to use here?  (It looks like the actual base class just returns the empty list.)

b2sdk/transfer/outbound/copy_source.py
```diff
@@ -13,8 +13,8 @@
 
 class CopySource(OutboundTransferSource):
     def __init__(self, file_id, offset=0, length=None):
-        if length is None and offset > 0:
-            raise ValueError('Cannot copy with non zero offset and unknown length')
+        if not length and offset > 0:
+            raise ValueError('Cannot copy with non zero offset and unknown or zero length')
```
bwbeach, line 17: Why?  Can't we copy a zero-length file?

b2sdk/transfer/outbound/copy_source.py
```diff
@@ -38,13 +38,15 @@ def is_copy(self):
         return True
 
     def get_bytes_range(self):
-        if self.length is None:
+        if not self.length:
```
bwbeach, line 41: here, too

bwbeach (commented):
This is a review of just the first commit.  (It's time for the next meeting, and I want the comments to be visible.)

Overall question: What are the semantics of copying from a hidden file?  It looks like the destination file gets deleted.  Is that what's expected?

bwbeach:
I went through the rest of the commits, and they look fine.

I do think there needs to be documentation on what happens when you sync from bucket to bucket.  I think it does not copy all the versions of files, just the most recent one.  And when in "delete" mode, when the source file is hidden, all versions of the destination file will be deleted.

b2sdk/transfer/outbound/copy_source.py
ppolewicz: We could, but we'd rather not. Uploads are A-class transactions (free), but `b2_copy_file` are C-class transactions (2500/day for free, then $0.004 per 1,000 calls), so since we are not going to save much bandwidth on transfer, it's better to re-upload the "whole" file.

b2sdk/sync/policy.py
ppolewicz: If the parent class will start returning something, we'd like this class (and all descendants) to honor that without modification of the inheriting classes

mlech-reef:
> Overall question: What are the semantics of copying from a hidden file? It looks like the destination file gets deleted. Is that what's expected?
1. By default the destination file is untouched.
2. If you pass `--delete` and there is no file on the source, the destination is deleted.
3. If you pass `--delete` and the file on the source is hidden, the destination file **is untouched**.

For me 1 and 2 are ok, but 3 is weird behavior but is not something that I introduced as it works exactly the same way when the destination is a local folder.

ppolewicz:
Please file a separate issue for that so that we may decide whether this is a bug that needs to be fixed for local folders too

b2sdk/bucket.py
mlech-reef: The first branch is using a simple copy manager and the second emerger created by @mzukowski-reef 
The problem is that emerger doesn't work when copying 0-length files and actually is not needed for such files

b2sdk/transfer/outbound/copy_source.py
mlech-reef: This is an emerger code. See my previous comment.

b2sdk/transfer/outbound/copy_source.py
mlech-reef: Same here

mlech-reef:
> I do think there needs to be documentation on what happens when you sync from bucket to bucket.
Where would you like to see such documentation? In CLI? I think that overall sync documentation in b2sdk is less informative comparing to CLI sync command description.

ppolewicz:
both, actually.

Also, we'll add sync of all versions in a separare PR

mlech-reef:
I've added a better docstring, updated the sphinx docs slightly, and created tickets for dealing with hidden files and sync of all versions (linked).

mlech-reef:
An update for the CLI: Backblaze/B2_Command_Line_Tool#667

b2sdk/sync/sync.py
```diff
@@ -96,6 +96,32 @@ class KeepOrDeleteMode(Enum):
...
+      It will sync only the latest versions of files.
+
+    By default, the synchronizer:
+
+    - Fails when the specified source directory doesn't exist or is empty.
+      (see ``allow_empty_source`` argument)
+    - Fails when the source is newer.
+      (see ``newer_file_mode`` argument)
+    - Doesn't delete a file if it's present on the destination but not on the source.
+      When the source is B2 and the file is hidden, the file on the destination won't be deleted
```
ppolewicz, line 118: I think it's a bug https://github.com/Backblaze/b2-sdk-python/issues/167 and we should not document it but we should fix it

b2sdk/sync/sync.py
```diff
@@ -96,6 +96,32 @@ class KeepOrDeleteMode(Enum):
 
 
 class Synchronizer(object):
+    """
+    Copies multiple files from source to destination.  Optionally
```
ppolewicz, line 100:
```suggestion
    Copies multiple "files" from source to destination.  Optionally
```
those are objects, really, but we call them files here.

b2sdk/sync/sync.py
```diff
@@ -96,6 +96,32 @@ class KeepOrDeleteMode(Enum):
 
 
 class Synchronizer(object):
+    """
```
ppolewicz, line 99: the description doesn't mention the fact that we only take the newest version when handling bucket-to-bucket transfer, and we don't take old versions. It should be mentioned.

doc/source/api/sync.rst
```diff
@@ -27,7 +27,7 @@ Following are the important optional arguments that can be provided while initia
...
 * ``compare_threshold``: It's the minimum size (in bytes)/modification time (in seconds) difference between source and destination files before we assume that it is new and replace.
 * ``newer_file_mode``: To identify whether to skip or replace if source is older. For possible values see :class:`b2sdk.v1.NewerFileSyncMode`. If you don't specify this the sync will raise :class:`b2sdk.v1.exception.DestFileNewer` in case any of the source file is older than destination.
-* ``keep_days_or_delete``: specify policy to keep or delete older files. For possible values see :class:`b2sdk.v1.KeepOrDeleteMode`. Default is `DO_NOTHING`.
+* ``keep_days_or_delete``: specify policy to keep or delete older files. For possible values see :class:`b2sdk.v1.KeepOrDeleteMode`. Default is `DO_NOTHING`. When the source is B2 and the file is hidden, the file on the destination won't be deleted even with :py:attr:`b2sdk.v1.KeepOrDeleteMode.DELETE`.
```
ppolewicz, line 30: whoah, don't add it to the interface - it's a bug!

b2sdk/sync/sync.py
mlech-reef: It is in line 109:
>    The synchronizer can copy files:
>
>    ...
>    - Between different folders in the same B2 bucket.
>      **It will sync only the latest versions of files.**

ppolewicz:
Codacy says:
test/unit/v1/test_sync.py
unnecessary `from enum import Enum`

apart from that I think we are ready to merge

mlech-reef:
> Codacy says:
> test/unit/v1/test_sync.py
> unnecessary `from enum import Enum`
> 
> apart from that I think we are ready to merge

Fixed.

ppolewicz (approved):

Issue #143, opened 2020-07-07
Getting File ID from name


selo2020:
Hi,

I know that this question was asked previously [here](https://github.com/Backblaze/b2-sdk-python/issues/122) but I had some further questions regarding this functionality. 

You previously mentioned that we could use `list_file_versions` to do this - which is fairly easy to do (just limit the fetch_count to 1). But my concern with that is that in terms of "API cost" it's more expensive than calling `get_file_id` - as `list_file_versions` is in the same class as list_file_names. 

So my question is, could you have a call which assumes the latest version for the file you want the information about? 

If this is resolved with some other solution then I would be happy to use that but I am just trying to reduce the number of API calls and I couldn't see any obvious alternative. 

Thanks

ppolewicz:
I get what you are saying, but there is no `get_file_id` in the list of b2 api available operations. If the server would expose such operation, we would expose it as well in the sdk interface, but as of now, there is not much b2-sdk-python can do (unless I misunderstood something?)

selo2020:
Ok that's cool. Is there any recommended way that we can get similar information without the requirement of a `list_file_versions` call? Or is that the only way right now? (or caching from a `list_file_names` call)

ppolewicz:
Before I answer that, could you please describe the entire operation that you are trying to perform? I bet it's not just to delete a file by name but actually something much bigger is going on. If I can understand that, it would let me to help you better.

selo2020:
I am trying to download a file by name but I want to find out some file metadata before doing so - crucially the size and modification date. Although this is only one part of it - I would also like to be able to get the metadata for a specific file (almost like a stat command). I would rather not cache the listings in memory either as there could be hundreds of thousands of files returned. 

I am aware that there are calls to be able to download a file but I am trying to perform this task without utilising local storage (as we cannot guarantee there is a large amount of storage available) or by putting it all in memory (potentially some files are >100GB in size). So I am trying to implement a range downloader to allow me to "consume" the file, but to do that I need to know the size in advance. Generally I know the file exists, but there could also be the instance where it does not.

Bascially, I am thinking of an operation akin to HeadObject in S3 where you can specify the Key and optionally a VersionId if you need to query a specific version.

ppolewicz:
Why are you checking the metadata of a file before downloading it? If the metadata doesn't match a condition, will you select an older version?

Are you downloading/consuming a single file or many files in part of one big operation?

[...]

PR #172, opened 2020-11-09
Implement bucket.get_file_info_by_id and bucket.get_file_info_by_name

ehossack:
This attempts to address https://github.com/Backblaze/b2-sdk-python/issues/143

This required edits through various layers of the api.

I found the test structure rather strange, and couldn't figure out why the test I wrote was failing, so if **the maintainers could have a look and help**, that would be swell.

Also, I included a commit which added type hints for the methods I touched, because it made the code easier to follow, and the way to do this for a codebase is to start incrementally, but I understand if the maintainers want to hold off adding type hints for now.

### DIFF
--- a/b2sdk/api.py
+++ b/b2sdk/api.py
@@ -8,6 +8,7 @@
 #
 ######################################################################
 
+from typing import Any, Dict
 from .bucket import Bucket, BucketFactory
 from .exception import NonExistentBucket, RestrictedBucket
 from .file_version import FileIdAndName
@@ -423,7 +424,7 @@ class B2Api(metaclass=B2TraceMeta):
         )
 
     # other
-    def get_file_info(self, file_id):
+    def get_file_info(self, file_id: str) -> Dict[str, Any]:
         """
         Legacy interface which just returns whatever remote API returns.
 
--- a/b2sdk/api.py
+++ b/b2sdk/api.py
@@ -429,8 +429,12 @@ class B2Api(metaclass=B2TraceMeta):
 
         .. todo::
             get_file_info() should return a File with .delete(), copy(), rename(), read() and so on
+
+        :param str file_id: the id of the file who's info will be retrieved.
+        :return: The parsed response
+        :rtype: dict
         """
-        return self.session.get_file_info(file_id)
+        return self.session.get_file_info_by_id(file_id)
 
     def check_bucket_restrictions(self, bucket_name):
         """
--- a/b2sdk/b2http.py
+++ b/b2sdk/b2http.py
@@ -51,11 +51,11 @@ def _translate_errors(fcn, post_params=None):
         response = fcn()
         if response.status_code not in [200, 206]:
             # Decode the error object returned by the service
-            error = json.loads(response.content.decode('utf-8'))
+            error = json.loads(response.content.decode('utf-8')) if response.content else {}
             raise interpret_b2_error(
-                int(error['status']),
-                error['code'],
-                error['message'],
+                int(error.get('status', response.status_code)),
+                error.get('code'),
+                error.get('message'),
                 response.headers,
                 post_params,
             )
@@ -291,18 +291,17 @@ class B2Http(object):
         :return: a dict that is the decoded JSON
         :rtype: dict
         """
-        # Make the headers we'll send by adding User-Agent to what
-        # the caller provided.  Make a copy before modifying.
-        headers = dict(headers)  # make copy before modifying
-        headers['User-Agent'] = self.user_agent
+        request_headers = {**headers, 'User-Agent': self.user_agent}
 
         # Do the HTTP POST.  This may retry, so each post needs to
         # rewind the data back to the beginning.
         def do_post():
             data.seek(0)
-            self._run_pre_request_hooks('POST', url, headers)
-            response = self.session.post(url, headers=headers, data=data, timeout=self.TIMEOUT)
-            self._run_post_request_hooks('POST', url, headers, response)
+            self._run_pre_request_hooks('POST', url, request_headers)
+            response = self.session.post(
+                url, headers=request_headers, data=data, timeout=self.TIMEOUT
+            )
+            self._run_post_request_hooks('POST', url, request_headers, response)
             return response
 
         try:
@@ -344,7 +343,7 @@ class B2Http(object):
 
     def get_content(self, url, headers, try_count=5):
         """
-        Fetche content from a URL.
+        Fetches content from a URL.
 
         Use like this:
 
@@ -366,20 +365,56 @@ class B2Http(object):
         :param int try_count: a number or retries
         :return: Context manager that returns an object that supports iter_content()
         """
-        # Make the headers we'll send by adding User-Agent to what
-        # the caller provided.  Make a copy before modifying.
-        headers = dict(headers)  # make copy before modifying
-        headers['User-Agent'] = self.user_agent
+        request_headers = {**headers, 'User-Agent': self.user_agent}
 
         # Do the HTTP GET.
         def do_get():
-            self._run_pre_request_hooks('GET', url, headers)
-            response = self.session.get(url, headers=headers, stream=True, timeout=self.TIMEOUT)
-            self._run_post_request_hooks('GET', url, headers, response)
+            self._run_pre_request_hooks('GET', url, request_headers)
+            response = self.session.get(
+                url, headers=request_headers, stream=True, timeout=self.TIMEOUT
+            )
+            self._run_post_request_hooks('GET', url, request_headers, response)
             return response
 
         response = _translate_and_retry(do_get, try_count, None)
         return ResponseContextManager(response)
+
+    def head_content(self, url, headers, try_count=5):
+        """
+        Does a HEAD instead of a GET for the URL.
+        The response's content is limited to the headers.
+
+        Use like this:
+
+        .. code-block:: python
+
+           try:
+               response_dict = b2_http.head_content(url, headers)
+               ...
+           except B2Error as e:
+               ...
+
+        The response object is only guaranteed to have:
+            - headers
+
+        :param str url: a URL to call
+        :param dict headers: headers to send
+        :param int try_count: a number or retries
+        :return: the decoded response
+        :rtype: dict
+        """
+        request_headers = {**headers, 'User-Agent': self.user_agent}
+
+        # Do the HTTP HEAD.
+        def do_head():
+            self._run_pre_request_hooks('HEAD', url, request_headers)
+            response = self.session.head(
+                url, headers=request_headers, stream=True, timeout=self.TIMEOUT
+            )
+            self._run_post_request_hooks('HEAD', url, request_headers, response)
+            return response
+
+        return _translate_and_retry(do_head, try_count, None)
 
     @classmethod
     def _get_user_agent(cls, user_agent_append):
--- a/b2sdk/b2http.py
+++ b/b2sdk/b2http.py
@@ -17,6 +17,8 @@ import arrow
 import requests
 import time
 
+from typing import Any, Dict
+
 from .exception import (
     B2Error, B2RequestTimeoutDuringUpload, BadDateFormat, BrokenPipe, B2ConnectionError,
     B2RequestTimeout, ClockSkew, ConnectionReset, interpret_b2_error, UnknownError, UnknownHost
@@ -379,7 +381,7 @@ class B2Http(object):
         response = _translate_and_retry(do_get, try_count, None)
         return ResponseContextManager(response)
 
-    def head_content(self, url, headers, try_count=5):
+    def head_content(self, url: str, headers: Dict[str, Any], try_count: int = 5) -> Dict[str, Any]:
         """
         Does a HEAD instead of a GET for the URL.
         The response's content is limited to the headers.
--- a/b2sdk/bucket.py
+++ b/b2sdk/bucket.py
@@ -171,7 +171,7 @@ class Bucket(metaclass=B2TraceMeta):
             url, download_dest, progress_listener, range_
         )
 
-    def get_file_info_by_id(self, file_id):
+    def get_file_info_by_id(self, file_id: str) -> FileVersionInfo:
         """
         Gets a file version's info by ID.
 
@@ -180,7 +180,7 @@ class Bucket(metaclass=B2TraceMeta):
         """
         return FileVersionInfoFactory.from_api_response(self.api.get_file_info(file_id))
 
-    def get_file_info_by_name(self, file_name):
+    def get_file_info_by_name(self, file_name: str) -> FileVersionInfo:
         """
         Gets a file version's info by its name.
 
--- a/b2sdk/bucket.py
+++ b/b2sdk/bucket.py
@@ -10,8 +10,8 @@
 
 import logging
 
-from .exception import UnrecognizedBucketType
-from .file_version import FileVersionInfoFactory
+from .exception import FileNotPresent, FileOrBucketNotFound, UnrecognizedBucketType
+from .file_version import FileVersionInfo, FileVersionInfoFactory
 from .progress import DoNothingProgressListener
 from .transfer.emerge.executor import AUTO_CONTENT_TYPE
 from .transfer.emerge.write_intent import WriteIntent
@@ -156,7 +156,7 @@ class Bucket(metaclass=B2TraceMeta):
 
             :ref:`Synchronizer <sync>`, a *high-performance* utility that synchronizes a local folder with a Bucket.
 
-        :param str file_id: a file ID
+        :param str file_name: a file name
         :param download_dest: an instance of the one of the following classes: \
         :class:`~b2sdk.v1.DownloadDestLocalFile`,\
         :class:`~b2sdk.v1.DownloadDestBytes`,\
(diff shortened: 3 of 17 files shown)


b2sdk/bucket.py
```diff
@@ -171,6 +171,26 @@ def download_file_by_name(self, file_name, download_dest, progress_listener=None
...
+
+    def get_file_info_by_name(self, file_name):
+        """
+        Gets a file version's info by its name.
+
+        :param str file_name: the name of the file info to retrieve.
+        :rtype: generator[b2sdk.v1.FileVersionInfo]
+        """
+        response = self.api.session.get_file_info_by_name(self.id_, file_name)
+        return FileVersionInfoFactory.from_api_response(response)
```
ehossack, line 192: My guess is actually this needs some coercion, so will test in eventually, but focusing on the unit tests first.

b2sdk/raw_api.py
```diff
@@ -143,7 +144,14 @@ def get_download_authorization(
         pass
 
     @abstractmethod
-    def get_file_info(self, api_url, account_auth_token, file_id):
+    def get_file_info_by_id(self, api_url: str, account_auth_token: str,
+                            file_id: str) -> Dict[str, Any]:
+        pass
+
+    @abstractmethod
+    def get_file_info_by_name(
```
mlech-reef, line 152: In `bucket.py` and `session.py` you use `bucket_id` but here you have `bucket_name`.

This is a direct cause of the unit tests failure:
```
self = <b2sdk.raw_simulator.RawSimulator object at 0x7f3fac713c50>, api_url = 'http://api.example.com', account_auth_token = 'auth_token_0', bucket_name = 'bucket_0', file_name = 'a'

    def get_file_info_by_name(self, api_url, account_auth_token, bucket_name, file_name):
>       bucket = self.bucket_name_to_bucket[bucket_name]
E       KeyError: 'bucket_0'

b2sdk/raw_simulator.py:942: KeyError
```

mlech-reef (commented):
The unit tests should be updated/extended for both `v0` and `v1` unless you use the newly introduced testing schema using pytest (see `test/unit/sync/` as an example).
The tests from `v0` and `v1` only run for that particular version and all others run for both.

b2sdk/bucket.py
```diff
@@ -171,6 +171,26 @@ def download_file_by_name(self, file_name, download_dest, progress_listener=None
             url, download_dest, progress_listener, range_
         )
 
+    def get_file_info_by_id(self, file_id: str) -> FileVersionInfo:
```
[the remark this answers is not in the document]
ehossack: Yeah, I agree it's kindof weird, but more consistent to keep it here as well? As a user I'd just intuitively expect a file operation to be associated with a bucket.

b2sdk/raw_api.py
```diff
@@ -407,9 +415,18 @@ def get_download_authorization(
...
         )
 
-    def get_file_info(self, api_url, account_auth_token, file_id):
+    def get_file_info_by_id(self, api_url: str, account_auth_token: str,
+                            file_id: str) -> Dict[str, Any]:
         return self._post_json(api_url, 'b2_get_file_info', account_auth_token, fileId=file_id)
 
+    def get_file_info_by_name(self, account_auth_token: str, bucket_name: str,
+                              file_name: str) -> Dict[str, Any]:
+        download_url = self.get_download_url_by_name(bucket_name, file_name)
```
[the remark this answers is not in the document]
ehossack: I swear I know how to code...

b2sdk/bucket.py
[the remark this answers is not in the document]
ppolewicz: We have one case like that already (with download_file_by_id) and we should follow the same pattern - implement the logic in B2Api and then from Bucket call the API method. Do not implement the same thing twice and do not only keep it in the Bucket (because someone might want to get the file info for a fileid without knowing which bucket it is in)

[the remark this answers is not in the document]
ehossack: Can you give a clearer direction here? I'm not exactly sure what is being asked.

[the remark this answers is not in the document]
ppolewicz: please move that method to B2Api and then make a single-line function in Bucket that will `return self.api.get_file_info_by_id(file_id)`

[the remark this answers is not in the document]
ehossack: Okay, that’s clear. We already have a get_file_info method there. Should we deprecate that one and proceed?

If not, I can just call that existing method.
If so, we could use the [deprecation](https://pypi.org/project/deprecation/) library? (I double checked and the licenses of it and its dependencies are compatible with this library - Apache or MIT)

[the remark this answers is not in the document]
ppolewicz: In this project deprecation is handled differently - we have those `from b2sdk.v1 import ...` imports which keep the old interface working.

Do we really need to change the name of the method though? I think it is ok to have `Bucket.get_file_info_by_id(file_id)` call `self.api.get_file_info(fileId)`. Here `Bucket.get_file_info_by_id` will ust be a convenient alias (like `download_file_by_id`)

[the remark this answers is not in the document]
mlech-reef: Imo, I would do the oposit, just to be more consequent with the current naming:
* `download_file_by_id` / `download_file_by_name`
* `get_bucket_by_id` / `get_bucket_by_name`

So I would add `get_file_info_by_id` and `get_file_info_by_name` and leave `get_file_info` as an alias.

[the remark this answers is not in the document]
mlech-reef: ... and deprecate it in v2.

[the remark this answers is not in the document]
ehossack: Okay I will leave as is then for now - i.e. use get_file_info 👍

The deprecation strategy in this project is a bit opaque (and I personally find the versioning code structure quite confusing), so that was my thought process there.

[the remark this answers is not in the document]
ppolewicz: It's a new approach. Thank you for motivating me to write down the rules :)

b2sdk/api.py
```diff
@@ -423,14 +424,18 @@ def list_keys(self, start_application_key_id=None):
         )
 
     # other
-    def get_file_info(self, file_id):
+    def get_file_info(self, file_id: str) -> Dict[str, Any]:
```
mlech-reef, line 427: Just add a `get_file_info_by_id` method as alias here as everywhere else we use that kind of scheme

test/unit/v1/test_bucket.py
```diff
@@ -241,6 +241,30 @@ def _make_file(self, file_id, file_name):
         return self.bucket.start_large_file(file_name, 'text/plain', {})
 
 
+class TestGetFileInfo(TestCaseWithBucket):
```
mlech-reef, line 244: Do the same for `v0`

b2sdk/raw_api.py
```diff
@@ -407,9 +416,19 @@ def get_download_authorization(
...
-    def get_file_info(self, api_url, account_auth_token, file_id):
+    def get_file_info_by_id(self, api_url: str, account_auth_token: str,
+                            file_id: str) -> Dict[str, Any]:
         return self._post_json(api_url, 'b2_get_file_info', account_auth_token, fileId=file_id)
 
+    def get_file_info_by_name(
+        self, download_url: str, account_auth_token: str, bucket_name: str, file_name: str
+    ) -> Dict[str, Any]:
+        download_url = self.get_download_url_by_name(download_url, bucket_name, file_name)
+        response = self.b2_http.head_url(
```
mlech-reef, line 427:
```Traceback (most recent call last):
  File "/home/maciek/Workspace/b2/b2-sdk-python/b2sdk/raw_api.py", line 775, in test_raw_api
    test_raw_api_helper(raw_api)
  File "/home/maciek/Workspace/b2/b2-sdk-python/b2sdk/raw_api.py", line 945, in test_raw_api_helper
    download_url, account_auth_token, bucket_name, file_name
  File "/home/maciek/Workspace/b2/b2-sdk-python/b2sdk/raw_api.py", line 427, in get_file_info_by_name
    response = self.b2_http.head_url(
AttributeError: 'B2Http' object has no attribute 'head_url'
```
Please run the integration test manually, as CI doesn't run it from forks

test/unit/v1/test_bucket.py
ehossack: Why would we do that? We're not backporting this to v0 are we?

b2sdk/api.py
ehossack: Based on the last conversation, I was under the impression we are not doing this. Can you explain?

b2sdk/raw_api.py
ehossack: What's required to run the integration test? I don't want to pay money to run a test.

b2sdk/api.py
mlech-reef: Ok, you're right. Let's keep it then

test/unit/v1/test_bucket.py
mlech-reef: It's backported automatically. This is how ApiVer works (the thing that is used here).
In short: `v0` imports everything from `v1` and if we need to preserve the old behavior in v0, we create wrappers there.
@ppolewicz can say more about that ;)

b2sdk/raw_api.py
mlech-reef: Good question. We have recently changed the CI and enabled integration tests, but for security reasons, the integration tests don't work from forks as they use GitHub secrets. We may use [pull_request_target](https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#pull_request_target) in the future to enable the integration tests from forks, but until then contributors can use free B2 accounts which are more than enough to run those tests. Most of the tests in B2 SDK use the simulator anyway

ehossack: On running the tests locally, I found out also the SDK exception handling doesn't handle 404 very well (as discovered when the integration tests failed due to a dependent test that hid the file.... those tests should definitely be split out and some error cases tested...).
So I added two exception types to ensure that we don't mis-interpret those cases.

b2sdk/exception.py
```diff
@@ -362,6 +366,16 @@ class NonExistentBucket(B2SimpleError):
     prefix = 'No such bucket'
 
 
+class FileOrBucketNotFound(B2SimpleError):
```
mlech-reef, line 369: Maybe? I think it'd be backward-compatible
```
      ResourceNotFound
            |
    FileOrBucketNotFound
    /                  \
NonExistentBucket    FileNotPresent
```

ehossack: Seems reasonable to me.

b2sdk/bucket.py
```diff
@@ -171,6 +171,37 @@ def download_file_by_name(self, file_name, download_dest, progress_listener=None
...
+                file_name=response_headers.get('x-bz-file-name'),
+                size=response_headers.get('content-length'),
+                content_type=response_headers.get('content-type'),
+                content_sha1=response_headers.get('x-bz-content-sha1'),
+                file_info=None,
+                upload_timestamp=response_headers.get('x-bz-upload-timestamp'),
+                action=None
+            )
+        except FileOrBucketNotFound:
+            raise FileNotPresent()
```
mlech-reef, line 203: Maybe pass some info like in `raw_simulator.py`:
```FileNotPresent("No file with name [{}] in bucket".format(file_name))```

File: b2sdk/exception.py [hunk 10]
```diff
@@ -358,7 +378,7 @@ class MissingPart(B2SimpleError):
     prefix = 'Part number has not been uploaded'
 
 
-class NonExistentBucket(B2SimpleError):
+class NonExistentBucket(FileOrBucketNotFound):
```
mlech-reef, line 381: Please override `__str__` as well

mlech-reef (approved):

b2sdk/raw_api.py
```diff
@@ -143,7 +147,14 @@ def get_download_authorization(
...
     @abstractmethod
-    def get_file_info(self, api_url, account_auth_token, file_id):
+    def get_file_info_by_id(self, api_url: str, account_auth_token: str,
+                            file_id: str) -> Dict[str, Any]:
+        pass
+
+    @abstractmethod
+    def get_file_info_by_name(
+        self, download_url: str, account_auth_token: str, bucket_name: str, file_name: str
+    ) -> CaseInsensitiveDict:
```
ppolewicz, line 157: this ties raw_api interface with `requests`, but soon we will move to `curl` and then having requests as a dependency in the interface will be awkward. Can we please use `Dict[str, str]` here to avoid the dependency in the interface?

b2sdk/raw_api.py
```diff
@@ -281,6 +292,7 @@ class B2RawApi(AbstractRawApi):
 
     def __init__(self, b2_http):
         self.b2_http = b2_http
+        self.logger = getLogger(__name__)
```
ppolewicz, line 295: Here you would potentially have multiple loggers with the same name - I'm not sure if this is a good idea. If we'd initialize multiple classes in this file, they would all need to run getLogger in their constructors to get access to the logger, keep a reference to it, a module-level function wouldn't have anywhere to keep it in state... That is why in `b2sdk` by convention we use module-scoped loggers that we define right after imports as `logger = logging.getLogger(__name__)`. If it is not a big problem, please follow that convention.

b2sdk/raw_api.py
```diff
@@ -290,7 +302,8 @@ def _post_json(self, base_url, api_name, auth, **params):
         :param auth: passed in Authorization header
         :param api_name: example: "b2_create_bucket"
         :param args: the rest of the parameters are passed to b2
-        :return:
+        :return: the decoded JSON response
+        :rtype: dict
```
ppolewicz, line 306: ` -> Dict[str, Any]:` in the method signature should be better (some IDEs support it etc)

b2sdk/raw_api.py
```diff
@@ -407,9 +420,23 @@ def get_download_authorization(
...
         )
 
-    def get_file_info(self, api_url, account_auth_token, file_id):
+    def get_file_info_by_id(self, api_url: str, account_auth_token: str,
+                            file_id: str) -> Dict[str, Any]:
         return self._post_json(api_url, 'b2_get_file_info', account_auth_token, fileId=file_id)
 
+    def get_file_info_by_name(
+        self, download_url: str, account_auth_token: str, bucket_name: str, file_name: str
+    ) -> CaseInsensitiveDict:
```
ppolewicz, line 429: requests used in interface (as described in the abstract class)

b2sdk/session.py
```diff
@@ -183,8 +186,11 @@ def get_download_authorization(self, bucket_id, file_name_prefix, valid_duration
             valid_duration_in_seconds
         )
 
-    def get_file_info(self, file_id):
-        return self._wrap_default_token(self.raw_api.get_file_info, file_id)
+    def get_file_info_by_id(self, file_id: str) -> Dict[str, Any]:
+        return self._wrap_default_token(self.raw_api.get_file_info_by_id, file_id)
+
+    def get_file_info_by_name(self, bucket_name: str, file_name: str) -> CaseInsensitiveDict:
```
ppolewicz, line 192: ` -> Dict[str, Any]:`

b2sdk/b2http.py
```diff
@@ -366,21 +367,57 @@ def get_content(self, url, headers, try_count=5):
...
+            response = self.session.get(
+                url, headers=request_headers, stream=True, timeout=self.TIMEOUT
+            )
+            self._run_post_request_hooks('GET', url, request_headers, response)
             return response
 
         response = _translate_and_retry(do_get, try_count, None)
         return ResponseContextManager(response)
 
+    def head_content(self, url: str, headers: Dict[str, Any], try_count: int = 5) -> Dict[str, Any]:
```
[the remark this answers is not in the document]
mlech-reef: When we implement other than `requests` transporting layer, it will be refactored anyway I guess.

b2sdk/bucket.py
```diff
@@ -171,6 +171,37 @@ def download_file_by_name(self, file_name, download_dest, progress_listener=None
             url, download_dest, progress_listener, range_
         )
 
+    def get_file_info_by_id(self, file_id: str) -> FileVersionInfo:
+        """
+        Gets a file version's info by ID.
+
+        :param str file_id: the id of the file who's info will be retrieved.
+        :rtype: generator[b2sdk.v1.FileVersionInfo]
```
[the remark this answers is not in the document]
mlech-reef: Yes, it's redundant now. The same as:

```
:param str file_id: the id of the file who's info will be retrieved.
```

can be replaced by:

```
:param file_id: the id of the file who's info will be retrieved.
```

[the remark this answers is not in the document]
mlech-reef: I compared it with `sphinx`.

Typing and reStructuredText:
image

Only typing:
image

For me, the second version is equally good.

I think we can also use https://github.com/agronholm/sphinx-autodoc-typehints to have the first version without types in `:param:` and `:rtype:`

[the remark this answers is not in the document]
mlech-reef: With `sphinx-autodoc-typehints`:

image

Still, we don't have links and python-style typing was removed. There is no option to tweak this behavior I'm afraid.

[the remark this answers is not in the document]
ppolewicz: yes - I thought have it already. Please add it.

b2sdk/b2http.py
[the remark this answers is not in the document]
ppolewicz: fair point

b2sdk/bucket.py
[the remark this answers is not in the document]
mlech-reef: Let's not add `sphinx-autodoc-typehints` and keep `reStructuredText` for now until we sort it out. Do you agree @ppolewicz?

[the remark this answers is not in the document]
ppolewicz: ok, separate PR, cool

b2sdk/raw_api.py
ehossack: Ah good. I hate this convention too. I copied it from somewhere in the project because I thought that was your convention, but it looks like that was from my `.venv` folder :P

ehossack:
I don't know why, but running `nox -s lint` locally does not yield this error for me.

ppolewicz (approved):

ppolewicz:
Good chunk of good work!

PR #174, opened 2020-12-01
Fix missing SyncReport import in documentation

cimm:
Quick documentation fix.

The synchronization example in the [documentation](https://b2-sdk-python.readthedocs.io/en/master/quick_start.html?highlight=SyncReport#synchronization) uses the `SyncReport` class but it's not imported so this can't work.


ppolewicz (approved):

ppolewicz:
Thanks for this @cimm, docs can always need some ❤️!

PR #180, opened 2021-01-10
If content SHA1 is provided, populate it in file info as large_file_sha1 for large files

sTywin:
B2 does not provide the contentSha1 value for large files, instead returning "none". This can cause confusion for many users [1], and the fact that the B2 CLI does not populate the large_file_sha1 file info entry is arguably a bug.

This implementation only adds the large_file_sha1 file info entry if it is not already present, and the content SHA1 is known in advance (i.e. it will not cause a SHA1 to be computed if it is not provided). This works with B2 CLI when specifying the "--sha1" option, which without this, would silently swallow the provided SHA1, which the user would likely not notice until _after_ the potentially lengthy upload was complete.

In addition, this implementation only applies for `emerge`, not `emerge_stream`, since it's not clear to me where the known SHA1 would be, assuming it were provided.
    
[1] https://github.com/Backblaze/B2_Command_Line_Tool/issues/604

### DIFF
--- a/b2sdk/transfer/emerge/emerger.py
+++ b/b2sdk/transfer/emerge/emerger.py
@@ -63,10 +63,12 @@ class Emerger(metaclass=B2TraceMetaAbstract):
         """
         planner = self.get_emerge_planner(recommended_upload_part_size=recommended_upload_part_size)
         emerge_plan = planner.get_emerge_plan(write_intents)
-        if emerge_plan.is_large_file() and 'large_file_sha1' not in file_info:
+        if emerge_plan.is_large_file():
             [write_intent] = write_intents
             if write_intent.outbound_source.is_sha1_known():
-                file_info['large_file_sha1'] = write_intent.outbound_source.get_content_sha1()
+                file_info = file_info or {}
+                if 'large_file_sha1' not in file_info:
+                    file_info['large_file_sha1'] = write_intent.outbound_source.get_content_sha1()
         return self.emerge_executor.execute_emerge_plan(
             emerge_plan,
             bucket_id,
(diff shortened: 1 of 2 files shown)


b2sdk/transfer/emerge/emerger.py
```diff
@@ -63,6 +63,12 @@ def emerge(
         """
         planner = self.get_emerge_planner(recommended_upload_part_size=recommended_upload_part_size)
         emerge_plan = planner.get_emerge_plan(write_intents)
+        if emerge_plan.is_large_file():
+            [write_intent] = write_intents
```
mzukowski-reef, line 67: there are multiple issues with this line:
1) `write_intents` suggests that it can contain multiple write intents - most usages in `Bucket` have single intent, but `Emerger` can handle multiple write intents - this is also documented clearly in doc string
2) `write_intents` can be an iterator - doc string says "iterable" - and if it is iterator - then `get_emerge_plan` is reading it to the end - and this line would fail

b2sdk/transfer/emerge/emerger.py
```diff
@@ -63,6 +63,12 @@ def emerge(
...
+            if write_intent.outbound_source.is_sha1_known():
```
mzukowski-reef, line 68: If you want to write it right you would need to check if all `emerge_plan.emerge_parts` `.part_definition` are instances of `UploadEmergePartDefinition` **and** if `.part_definition.upload_source` is the same object in every part - if the answer is yes - then you can set `large_file_sha1` is it is known for this **common** `upload_source`

mzukowski-reef (changes requested):

PR #181, opened 2021-01-16
Method to compare credentials in account_info

mpnowacki-reef:
Method needed for checking whether new authorization data and cached authorization data match or not. Needed for authorizing via environment variables. I did not write any tests because I wasn't sure if it's a good idea to put it here - maybe it should be left in CLI.


ppolewicz:
mistargeted PR

PR #184, opened 2021-01-21
Method to compare key in account_info

mpnowacki-reef:
Method for checking whether new authorization key and cached authorization key match or not (it compares the realm as well). Needed for authorizing via environment variables.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+* Add `is_same_key` method to `AccountInfo`.
+
+### Fixed
+* Remove unnecessary cache clearing when reauthorizing with the same key
+
 ## [1.3.0] - 2021-01-13
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 * Add an ability to provide `bucket_id` filter parameter for `list_buckets`
-* Add `is_same_key` method to `AccountInfo`.
+* Add `is_same_key` method to `AccountInfo`
 
 ### Fixed
 * Fix docs autogen
@@ -22,7 +22,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 * Fix missing import in the synchronization example
-* Use setuptools-scm for versioning
+* Use `setuptools-scm` for versioning
+* Clean up CI steps
 
 ## [1.2.0] - 2020-11-03
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,8 +6,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-### Added
+### Changed
+* Add an ability to provide `bucket_id` filter parameter for `list_buckets`
 * Add `is_same_key` method to `AccountInfo`.
+
+### Fixed
+* Fix docs autogen
 
 ## [1.3.0] - 2021-01-13
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,9 +9,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 * Add `is_same_key` method to `AccountInfo`.
 
-### Fixed
-* Remove unnecessary cache clearing when reauthorizing with the same key
-
 ## [1.3.0] - 2021-01-13
 
 ### Added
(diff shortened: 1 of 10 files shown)


mlech-reef (approved):

test/unit/v0/test_account_info.py
```diff
@@ -184,8 +201,11 @@ def _test_account_info(self, check_persistence):
             self.assertEqual('account_auth', info2.get_account_auth_token())
             self.assertEqual('api_url', info2.get_api_url())
             self.assertEqual('app_key', info2.get_application_key())
+            self.assertEqual('key_id', info2.get_application_key_id())
             self.assertEqual('realm', info2.get_realm())
             self.assertEqual(100, info2.get_minimum_part_size())
+            self.assertTrue(info2.is_same_key('key_id', 'realm'))
+            self.assertFalse(info2.is_same_key('key_id', 'another_realm'))
```
ppolewicz, line 208:
```suggestion
            self.assertFalse(info2.is_same_key('key_id', 'another_realm'))
            self.assertFalse(info2.is_same_key('another_key_id', 'realm'))
```

test/unit/v1/test_account_info.py
```diff
@@ -180,8 +197,11 @@ def _test_account_info(self, check_persistence):
             self.assertEqual('account_auth', info2.get_account_auth_token())
             self.assertEqual('api_url', info2.get_api_url())
             self.assertEqual('app_key', info2.get_application_key())
+            self.assertEqual('key_id', info2.get_application_key_id())
             self.assertEqual('realm', info2.get_realm())
             self.assertEqual(100, info2.get_minimum_part_size())
+            self.assertTrue(info2.is_same_key('key_id', 'realm'))
+            self.assertFalse(info2.is_same_key('key_id', 'another_realm'))
```
ppolewicz, line 204:
```suggestion
            self.assertFalse(info2.is_same_key('key_id', 'another_realm'))
            self.assertFalse(info2.is_same_key('another_key_id', 'realm'))
```

b2sdk/session.py
```diff
@@ -100,12 +100,7 @@ def authorize_account(self, realm, application_key_id, application_key):
         :param str application_key: user's :term:`application key`
         """
         # Clean up any previous account info if it was for a different account.
```
ppolewicz, line 102: this comment is no longer accurate. Please remove it altogether.

b2sdk/session.py
```diff
@@ -100,12 +100,7 @@ def authorize_account(self, realm, application_key_id, application_key):
...
-        try:
-            old_account_id = self.account_info.get_account_id()
-            old_realm = self.account_info.get_realm()
-            if application_key_id != old_account_id or realm != old_realm:
-                self.cache.clear()
-        except MissingAccountData:
+        if not self.account_info.is_same_key(application_key_id, realm):
```
ppolewicz, line 103: This is better than it used to be, but still not good. It used to only work well for master key, now it works for the same key (nonmaster1 -> nonmaster1), but still it doesn't work between two non-master keys and there is actually a regression for switching between non-master and master key (the cache gets cleared in this particular case now, while it was not cleared in the past).

In order to actually fix it you need to compare old_account_id with the new account id (that we'll only find out if we manage to do the remote call).

The cache should not be cleared if we are moving between two different keys from the same account. Also, it needs more tests.

I suggest that you split this PR - ship `is_same_key()` first so that you can proceed with the CLI part, then open a new PR to solve the cache clearing issue.

mlech-reef:
This is in conflict with #142. Please post there a message to close the PR when you deliver it before #142

b2sdk/account_info/abstract.py
```diff
@@ -10,6 +10,7 @@
 
 from abc import abstractmethod
 
+from . import exception
```
ppolewicz, line 13: do not use relative imports and absolute imports together. The style of this project is to use absolute imports, so use that.

test/unit/v0/test_account_info.py
```diff
@@ -184,8 +201,12 @@ def _test_account_info(self, check_persistence):
...
+            self.assertFalse(info2.is_same_key('another_key_id', 'realm'))
```
ppolewicz, line 209:
```suggestion
            self.assertFalse(info2.is_same_key('another_key_id', 'realm'))
            self.assertFalse(info2.is_same_key('another_key_id', 'another_realm'))
```
The bug this additional test case protects from would be pretty weird, but I'd write it

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 * Add an ability to provide `bucket_id` filter parameter for `list_buckets`
+* Add `is_same_key` method to `AccountInfo`.
```
ppolewicz, line 11:
```suggestion
* Add `is_same_key` method to `AccountInfo`
```

mlech-reef:
Looks good. Please just remember to create a ticket regarding:

> I suggest that you split this PR - ship is_same_key() first so that you can proceed with the CLI part, then open a new PR to solve the cache clearing issue.

PR #186, opened 2021-02-04
Fix save buckets with the same name

mandricmihai:
Fixes https://github.com/Backblaze/B2_Command_Line_Tool/issues/567
When deleteing the bucket the id was used but that is a problem.
Call save() first time, bucket is added in table.
Call save() second time,  the first bucket is not deleted because the id is diferent and now we have two buckets with the same name but different id's and one of them is from a deleted bucket (invalid)
From the b2 cli the user can't call save() twice, it will call delete() and then save().
delete() [doesn't invalidate the cache ](https://github.com/Backblaze/b2-sdk-python/blob/4df28a4f896ed1e14b4e50f9ef2a94c58801337d/b2sdk/api.py#L275)

image


mandricmihai:
This is not the right way to solve this issue, pr will be closed.

PR #202, opened 2021-03-04
initial dependabot config

ooola:


mlech-reef:
Looks nice. I have one question. I'm wondering how dependabot's pip package ecosystem is looking for dependencies.
I know that it looks in `requirements.txt` where we have runtime dependencies, however, we store other types of python dependencies in other places:

* `setyp.py` for documentation - as python extras - although we don't have any version classifiers there and we always use the latest, so it's not an issue:
```python
extras_require={
        'doc':
            [
                'sphinx', 'sphinx-autobuild', 'sphinx_rtd_theme', 'sphinxcontrib-plantuml',
                'sadisplay'
            ],
    },
```

* `noxfile.py` for development - as python lists:
```python
REQUIREMENTS_FORMAT = ['yapf==0.27']                                                                                                                                                                               
REQUIREMENTS_LINT = ['yapf==0.27', 'pyflakes==2.2.0', 'pytest==6.1.1', 'liccheck==0.4.7']                                                                                                                          
REQUIREMENTS_TEST = ['pytest==6.1.1', 'pytest-cov==2.10.1', 'pytest-mock==3.3.1']                                                                                                                                  
REQUIREMENTS_BUILD = ['setuptools>=20.2']
```

I'm pretty sure that the dependabot won't help us with the second point and would be really nice if it could.

We could move the dependencies to e.g.:
`requirements-format.txt` or `requirements/format.txt` for `REQUIREMENTS_FORMAT`, etc.
but I'm not sure if it's smart enough to properly find them. Do you know anything how it works?

ooola:
I don't think dependabot will look at dependencies in the noxfile. However if they're only development dependencies they're not as important for end users of the tool. Note that dependabot can be configured for both `production` and `development` but I think it only works with pip. 

The python dependabot configuration is documented here - https://docs.github.com/en/github/administering-a-repository/configuration-options-for-dependency-updates#python-index

mlech-reef:
Ok, so let's skip development dependencies for now. I'm wondering how `production` and `development` work for pip as in python world there is no one official way of storing dependencies and handling environments.

mlech-reef (approved):

PR #220, opened 2021-04-19
add __slots__ and __eq__ to FileVersionInfo

mpnowacki-reef:
reduce memory footprint of `FileVersionInfo` in preparation for SSE-C sync support needing to reference a full object if there is a chance that `SyncEncryptionSettingProvider` may need to read it

### DIFF
--- a/b2sdk/file_version.py
+++ b/b2sdk/file_version.py
@@ -107,10 +107,9 @@ class FileVersionInfo(object):
         return cls.LS_ENTRY_TEMPLATE % ('-', '-', '-', '-', 0, name)
 
     def __eq__(self, other):
-        if not isinstance(other, self.__class__):
-            return False
+        sentry = object()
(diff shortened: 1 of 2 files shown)


b2sdk/file_version.py
```diff
@@ -101,6 +106,14 @@ def format_folder_ls_entry(cls, name):
         """ legacy method, to be removed in v2: formats a `ls` "folder" consistently with format_ls_entry() """
         return cls.LS_ENTRY_TEMPLATE % ('-', '-', '-', '-', 0, name)
 
+    def __eq__(self, other):
+        if not isinstance(other, self.__class__):
+            return False
```
ppolewicz, line 111:
```suggestion
        if not isinstance(other, self.__class__):
            return False
```
I don't think we should check this. If it looks like a duck and quacks like a duck, lets treat it as a duck.

Issue #189, opened 2021-02-11
B2Api.get_bucket_by_id should return complete bucket as get_bucket_by_name does


mlech-reef:

mpnowacki-reef:
This feature is needed for projects like https://github.com/Backblaze/terraform-provider-b2 which implement it by themselves at the moment.

PR #229, opened 2021-04-28
Make B2Api.get_bucket_by_id return populated bucket objects in v2

mpnowacki-reef:
solves https://github.com/Backblaze/b2-sdk-python/issues/189

replaces https://github.com/reef-technologies/b2-sdk-python/pull/136


b2sdk/api.py
```diff
@@ -462,19 +478,32 @@ def get_file_info(self, file_id: str) -> Dict[str, Any]:
...
+        If not, it raises a :py:exc:`b2sdk.v1.exception.RestrictedBucket` error.
+
+        :raises b2sdk.v1.exception.RestrictedBucket: if the account is not allowed to use this bucket
+        """
+        self._check_bucket_restrictions('bucketId', bucket_id)
+
+    def _check_bucket_restrictions(self, key, value):
         allowed = self.account_info.get_allowed()
-        allowed_bucket_name = allowed['bucketName']
+        allowed_bucket_identifier = allowed[key]
```
ppolewicz, line 505: this kind of sounds like it could explode if I have a bucket with `name=bucket1, id=abcde` and another bucket with `name=abcde, id=bcdef`

ppolewicz (approved):

PR #233, opened 2021-04-29
refactored AccountInfo tests to single file using pytest

mpnowacki-reef:
replaces https://github.com/reef-technologies/b2-sdk-python/pull/144


test/unit/account_info/test_account_info.py
```diff
@@ -61,3 +71,268 @@ def test_s3_api_url(self, s3_api_url):
...
+            assert info2.is_same_key('key_id', 'realm')
+            assert not info2.is_same_key('key_id', 'another_realm')
+            assert not info2.is_same_key('another_key_id', 'realm')
+            assert not info2.is_same_key('another_key_id', 'another_realm')
+
+    @pytest.mark.apiver(from_ver=2)
+    def _test_account_info_v2(self):
+        account_info = self._make_info()
+        account_info.set_auth_data(
+            account_id='account_id',
```
ppolewicz, line 243: if the order of arguments (especially optional) is accidentally messed up after v2 release (even in v2+N in the future), then this test is going to pass, because it escapes the verification of the order of arguments. On the other hand, if we switch it to positional arguments and one argument is renamed accidentally, then the test will pass but invoking the function with its keyword arguments is going to fail.

I think we should call it twice, once with positional arguments and once with named, to be sure we did not mess up the external interface... but that goes for everything, not just for this method. Not even for b2sdk external interface, but for backward compatibility of every module (not just a python module, java etc has the same problem). It might be a good idea to think about this to harden the interface somehow. Lets discuss it separately though (not in this PR).

ppolewicz (approved):

PR #234, opened 2021-04-29
Minimum part size refactor

mpnowacki-reef:
replaces https://github.com/reef-technologies/b2-sdk-python/pull/143/


b2sdk/v1/session.py
```diff
@@ -0,0 +1,48 @@
...
+#
+# License https://www.backblaze.com/using_b2_code.html
+#
+######################################################################
+
+from b2sdk import _v2 as v2
+from .account_info import SqliteAccountInfo
+
+
+class B2Session(v2.B2Session):
```
ppolewicz, line 15: add a comment. Why are we doing this? `minimum_part_size=response['recommendedPartSize'],` is the critical piece, I think, but I'd rather read it from a comment and not from the code

mpnowacki-reef: you're right, I'm a strong supporter of documentaing reasons for apiver overrides, this one definetely needs a comment.

ppolewicz (approved):

PR #235, opened 2021-04-30
Add support for XDG_CONFIG_HOME

mpnowacki-reef:
replaces https://github.com/Backblaze/b2-sdk-python/pull/129


mzukowski-reef (approved):

ppolewicz (approved):
it's beautiful

ppolewicz:
on a side note, @mpnowacki-reef shoudln't we schedule removal of everything that starts with `os.path.` in favor of `pathlib`? Maybe add that to the list of things we can do when dropping 3.5?

mpnowacki-reef:
@ppolewicz Yes we could, but pathlib was introduced in 3.4, so i guess there's no need for it to happen only when dropping 3.5, right?

ppolewicz:
that is true, but we missed it back then, so now we track it in https://github.com/Backblaze/b2-sdk-python/issues/239 to not forget

PR #247, opened 2021-05-20
Fix crash when Content-Range header is missing

ppolewicz:
The server should return it if the range contains the whole file, but apparently it does not do that and that causes crashes which `b2-sdk-python` can just avoid with no consequences.

### DIFF
--- a/b2sdk/transfer/inbound/download_manager.py
+++ b/b2sdk/transfer/inbound/download_manager.py
@@ -19,7 +19,6 @@ from b2sdk.exception import (
     ChecksumMismatch,
     InvalidRange,
     TruncatedOutput,
-    UnexpectedCloudBehaviour,
 )
 from b2sdk.raw_api import SRC_LAST_MODIFIED_MILLIS
 from b2sdk.utils import B2TraceMetaAbstract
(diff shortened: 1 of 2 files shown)


b2sdk/transfer/inbound/download_manager.py
```diff
@@ -93,8 +92,7 @@ def download_file_from_url(
         ) as response:
             metadata = FileMetadata.from_response(response)
             if range_ is not None:
-                if 'Content-Range' not in response.headers:
-                    raise UnexpectedCloudBehaviour('Content-Range header was expected')
+                # 2021-05-20: unfortunately for a read of a complete object server does not return the 'Content-Range' header
```
mpnowacki-reef, line 95: well actually, we can determine whether server responded with the whole file or not by http status: 200 = full file and no `Content-Range` header; 206 = part of the file and `Content-Range` present

ppolewicz: B2Http response does not guarantee `.status_code` and I don't want to add it right now because we are not sure how well it is going to work with curl later, while having or not having this header doesn't really have that much impact. I don't think it's worth to do this well, even if it's possible. Do you think we should put the work in to keep the exception raising working?

mpnowacki-reef (approved):

b2sdk/transfer/inbound/download_manager.py
ppolewicz: we discussed it privately and determined it's not worth the effort

ericjding:
@ppolewicz just browsing some PRs and noticed this one; Content-Range doesn't seem to make sense (or be specified by RFC-7233) for anything other than a 206 (Partial Content) or 416 (Range Not Satisfiable) response, and even then, it's not strictly required by the RFC. Am I missing something?

ppolewicz:
It was a check that we didn't really need - you are correct, if it's a 200 OK and not 206, the header doesn't need to be there even if we requested a range

PR #253, opened 2021-05-29
Fix an issue calling incorrect api

ehossack:
This commit fixes a typo in calling the incorrect restrictions api (v1 version)

### DIFF
--- a/test/unit/v_all/test_api.py
+++ b/test/unit/v_all/test_api.py
@@ -61,3 +61,17 @@ class TestApi(TestBase):
         self.cache.save_bucket(Bucket(api=self.api, name='bucket_name', id_='bucket_id'))
         read_bucket = self.api.get_bucket_by_id('bucket_id')
         assert read_bucket.name == 'bucket_name'
+
+    def test_get_download_url_for_file_name(self):
+        self._authorize_account()
+
+        download_url = self.api.get_download_url_for_file_name('bucket1', 'some-file.txt')
+
+        assert download_url == 'http://download.example.com/file/bucket1/some-file.txt'
+
+    def test_get_download_url_for_fileid(self):
+        self._authorize_account()
+
+        download_url = self.api.get_download_url_for_fileid('file-id')
+
+        assert download_url == 'http://download.example.com/b2api/v2/b2_download_file_by_id?fileId=file-id'
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Refactored `FileVersionInfo` to `FileVersion`
 * `ScanPoliciesManager` exclusion interface changed
 
+### Fixed
+* Fix call to incorrect internal api in `B2Api.get_download_url_for_file_name`
+
 ## [1.8.0] - 2021-05-21
 
 ### Added
--- a/b2sdk/api.py
+++ b/b2sdk/api.py
@@ -434,7 +434,7 @@ class B2Api(metaclass=B2TraceMeta):
         :param str bucket_name: a bucket name
         :param str file_name: a file name
         """
-        self.check_bucket_restrictions(bucket_name)
+        self.check_bucket_name_restrictions(bucket_name)
         return '%s/file/%s/%s' % (
             self.account_info.get_download_url(), bucket_name, b2_url_encode(file_name)
         )


ehossack:
I'm surprised there's no tests for the `get_download_url_for_file*` apis.

I'm hoping this is grounds for a 1.8.1 release as it breaks the download url api.

Additionally, some guidance (outside of this PR, updating your own docs) on how to have all the available versions of python for nox would be appreciated. For example macOS no longer supports 3.6.x python versions.

ppolewicz (approved):

ppolewicz:
I think `nox` used `pyenv` to install all the python versions it wanted - I didn't install anything by myself and a full suite runs well on my box (except 3.5?), so I think the comment which suggests that you need to install all python versions on your OS by yourself is not accurate.

It would normally be important enough for a 1.8.1 release, but I think we'll have a 1.9.0 early next week with all the sync changes and v2 support anyway.

@mlech-reef could you please confirm what the situation is with installing python versions for the full nox suite to run tests locally? (we shouldn't merge the comment in if it's inaccurate)

ppolewicz:
(I had to approve the PR for the tests to run)

ehossack:
> I think `nox` used `pyenv` to install all the python versions it wanted - I didn't install anything by myself and a full suite runs well on my box (except 3.5?), so I think the comment which suggests that you need to install all python versions on your OS by yourself is not accurate.
> 
> It would normally be important enough for a 1.8.1 release, but I think we'll have a 1.9.0 early next week with all the sync changes and v2 support anyway.
> 
> @mlech-reef could you please confirm what the situation is with installing python versions for the full nox suite to run tests locally? (we shouldn't merge the comment in if it's inaccurate)

Makes sense if the comment is innacurate. I have `pyenv` configured and running any of the `nox` commands failed, and I couldn't get nox to figure out how to pick up my installed python versions - hence the `"which python"` in there. 
Also python 3.5 and 3.6 fail to install on my system (macOS 11) due to some limitations, just to keep that in mind for tests/prs. For my own project I just run all the commands in a docker container...

ppolewicz:
Maybe it's a bug on macos, lets see what @mlech-reef will have to say about it

mlech-reef:
I don't think `nox` installs python on your own. It never happens for me :) It uses versions from `['3.5', '3.6', '3.7', '3.8', '3.9']` by default. That list could be overridden via `PYTHON_VERSIONS` environment variable (it is documented in the developer guide). If the version is not available on the PATH, it will fail, e.g.:
image

If you have properly configured `global` versions in `pyenv` like:
image
or `local`, e.g.:
image

everything should work fine.

ehossack:
So the comment itself is correct. And CI provides a backup.

test/unit/v_all/test_api.py
```diff
@@ -61,3 +61,19 @@ def test_get_bucket_by_id_v2(self):
         self.cache.save_bucket(Bucket(api=self.api, name='bucket_name', id_='bucket_id'))
         read_bucket = self.api.get_bucket_by_id('bucket_id')
         assert read_bucket.name == 'bucket_name'
+
+    @pytest.mark.apiver(from_ver=0)
```
mlech-reef, line 65: It is redundant

test/unit/v_all/test_api.py
```diff
@@ -61,3 +61,19 @@ def test_get_bucket_by_id_v2(self):
...
+    def test_get_download_url_for_file_name(self):
+        self._authorize_account()
+
+        download_url = self.api.get_download_url_for_file_name('bucket1', 'some-file.txt')
+
+        assert download_url == 'http://download.example.com/file/bucket1/some-file.txt'
+
+    @pytest.mark.apiver(from_ver=0)
```
mlech-reef, line 73: It is redundant

mlech-reef (changes requested):

test/unit/v_all/test_api.py
ehossack: Is consistency across the tests not better than having some tests that have the annotation and some not? Happy to change it but as someone coming in to the code I would wonder why some have and some don't

ehossack:
Moved the tangential conversation around local setup to #255

test/unit/v_all/test_api.py
mlech-reef: Why marking all the tests with some defaults? The reason for having this mark is to be able to easily list/run only the tests that differ across apiver versions with e.g. `nox -s unit-3.9 -- --co -m apiver`

ppolewicz: I agree that we should remove it. It's clear to me that the decorator should only be used on tests which should execute for a limited subset of apiver interface versions.

ppolewicz:
ok, my mistake, I was under the impression that nox will install environments for itself

mlech-reef (approved):

test/unit/v_all/test_api.py
ehossack: Fair enough. I already updated the PR to remove it - but sounds like it was just my confusion :)

mlech-reef: Already approved :)

ppolewicz (approved):

ppolewicz:
Great contribution, added a test. I think it's not the first thing @ehossack contributed, right? Perhaps it's time for some more ambitious work we could do together?

ehossack:
> Great contribution, added a test. I think it's not the first thing @ehossack contributed, right? Perhaps it's time for some more ambitious work we could do together?

For now I will have to stay with the smaller contributions to save my time!

ehossack:
P.S. When can we expect 1.9.0 if no 1.8.1? 😛

ppolewicz:
It's very close now. @mpnowacki-reef do we have anything left for 1.9.0? A couple new options for sync I think?

Issue #196, opened 2021-02-19
Change ScanPoliciesManager interface (see TODOs)


mlech-reef:
`ScanPoliciesManager.should_exclude_file` should accept `File` object, not string
`ScanPoliciesManager.should_exclude_directory` should accept `AbstractFolder` object, not string

mpnowacki-reef:
@mlech-reef Why is this needed for V2?

mlech-reef:
It breaks compatibility.

mpnowacki-reef:
@ppolewicz In light of 

> change all path arguments everywhere (internally too) to use pathlib instead of strings

from https://github.com/Backblaze/b2-sdk-python/issues/139

should `should_exclude_file` and `should_exclude_directory ` indeed use `File` and `Folder` and not `pathlib.Path` ?

ppolewicz:
I think the problem here is that `should_exclude_file` and `should_exclude_directory` can have b2:// `File` / `Folder` passed into them and `pathlib.Path` doesn't support b2:// resources

mpnowacki-reef:
The way `ScanPoliciesManager.should_exclude_file` and `ScanPoliciesManager.should_exclude_directory` work right now is they compare B2 paths to sets of regexes - even though it's not documented, they never operate on local paths. Even when invoked in LocalFolder, a would-be-B2-path of a dir or a file is constructed and passed to the appropriate ``ScanPoliciesManager.should_exclude_*` method. I'm not sure it would be beneficial to create "possibly-existsing-b2-file" objects when scanning local folders just to match their paths against regexes.

@mlech-reef Is this interface change somehow needed elsewhere? Or do you need access to all B2File versions (carried by B2File) in your implementation of `ScanPoliciesManager` ?

mpnowacki-reef:
However, changing the interface to have the following methods:

 * `should_exclude_local_file` accepting a `LocalFile` object
 * `should_exclude_b2_file` accepting a `B2File` object
 * `should_exclude_local_folder` accepting a `LocalFolder` object
 * `should_exclude_b2_floder` accepting a `B2Folder` object
 
 and performing the translations internally would make for a clean, extensible API. The problem is that `ScanPoliciesManager` objects would need access to `local_dir` and `b2_dir` to perform translations accurately. Or maybe a subdelegate of `ScanPoliciesManager`, a policy manager aware of the particular sync context?

mlech-reef:
[...]

PR #254, opened 2021-05-29
FileVersionInfo refactored to FileVersion, ScanPoliciesManager interface refactored, several other improvements

mpnowacki-reef:
solves https://github.com/Backblaze/b2-sdk-python/issues/196


ppolewicz (approved):
This has been incrementally reviewed in 3 parts, 2 people each part.

We have added a couple of interface changes to the scope in anticipation of a large impact feature which is going to happen soon, to avoid another apiver increase.

Issue #195, opened 2021-02-19
Change SyncReport according to the deprecation TODOs


mlech-reef:
Check TODOs in `sync/report.py` for details

mpnowacki-reef:
I don't think this is a part of the public API, so such breaking changes can be done aside from apiver release

mlech-reef:
`SyncReport` is part of the public API

PR #270, opened 2021-06-21
Several improvements that are necessary for v2

mpnowacki-reef:
Solves, amongst others: https://github.com/Backblaze/b2-sdk-python/issues/195


CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * `B2RawApi` renamed to `B2RawHTTPApi`
 * `B2HTTP` tests are now common
 * `B2HttpApiConfig` class introduced to provide parameters like `user_agent_append` to `B2Api` without using internal classes in v2
+* `Bucket.update` return s a `Bucket` object in v2
```
ppolewicz, line 15: extra space

b2sdk/bucket.py
```diff
@@ -115,17 +115,16 @@ def set_type(self, bucket_type):
...
+        lifecycle_rules: Optional[dict] = None,
+        if_revision_is: Optional[int] = None,
         default_server_side_encryption: Optional[EncryptionSetting] = None,
         default_retention: Optional[BucketRetentionSetting] = None,
     ):
         """
         Update various bucket parameters.
-        For legacy reasons in apiver v1 it returns whatever server returned on b2_update_bucket call, v2 will change that.
 
         :param str bucket_type: a bucket type
```
ppolewicz, line 129: maybe worth mentioning allPublic and allPrivate here, since it's not easily available anywhere near here

b2sdk/bucket.py
```diff
@@ -136,16 +135,19 @@ def update(
...
-            account_id,
-            self.id_,
-            bucket_type=bucket_type,
-            bucket_info=bucket_info,
-            cors_rules=cors_rules,
-            lifecycle_rules=lifecycle_rules,
-            if_revision_is=if_revision_is,
-            default_server_side_encryption=default_server_side_encryption,
-            default_retention=default_retention,
+        return BucketFactory.from_api_bucket_dict(
```
ppolewicz: by convention it should be `self.BUCKET_FACTORY.from_api_bucket_dict(`

test/unit/bucket/test_bucket.py
```diff
@@ -828,6 +828,110 @@ def _make_file(self, bucket=None):
...
+                'lifecycle_rules': {
+                    'life': 'is life'
+                },
+                'options_set':
+                    set(),
+                'default_server_side_encryption':
+                    SSE_B2_AES,
+                'default_retention':
+                    BucketRetentionSetting(RetentionMode.COMPLIANCE, RetentionPeriod(years=7)),
+            }.items():
```
ppolewicz: this really needs `# yapf: disable` and manual formatting

b2sdk/file_version.py
```diff
@@ -206,7 +206,7 @@ def __init__(
         content_disposition: Optional[str],
         content_length: int,
         content_language: Optional[str],
-        expires: Optional[datetime],
+        expires: Optional[str],
```
ppolewicz, line 215: why?

b2sdk/api.py
```diff
@@ -440,34 +442,54 @@ def create_key(
             valid_duration_seconds=valid_duration_seconds,
             bucket_id=bucket_id,
             name_prefix=name_prefix
+            # As of 17.06.2021 the server does not accept an 'options' argument
```
ppolewicz, line 445: that's because it's a read-only property, You can use it to see if the bucket is S3-capable or not (and I guess legalhold-capable nowadays?)

b2sdk/v1/api.py
```diff
@@ -27,6 +27,8 @@
 # and to retain old style download_file_by_id signature (allowing for the new one as well) and exception
 # and to retain old style get_file_info return type
 # and to accept old-style raw_api argument
+# and to retain old style download_file_by_id signature (allowing for the new one as well) and exception
```
ppolewicz, line 30: this here line 30 is a clone of line 27

b2sdk/v1/api.py
```diff
@@ -169,3 +171,34 @@ def download_file_by_id(
                 raise
         else:
             return downloaded_file
+
+    def list_keys(self, start_application_key_id=None) -> dict:
+        """
+        List application keys.
+
+        :param start_application_key_id: an :term:`application key ID` to start from or ``None`` to start from the beginning
```
ppolewicz, line 179: it should be noted that only a max of `self.DEFAULT_LIST_KEY_COUNT` is going to be returned (unlike v2 which returns all keys, unless you stop iterating, I guess?)

b2sdk/exception.py
```diff
@@ -485,6 +485,10 @@ def __str__(self):
         return "%s cannot be used as default for a bucket." % (self.encryption_mode,)
 
 
+class CopyArgumentsMismatch(B2Error):
```
ppolewicz, line 488: no common abstract exception for user input errors?

test/unit/sync/test_sync_report.py
```diff
@@ -10,12 +10,10 @@
...
 from unittest.mock import MagicMock
 
-from ..test_base import TestBase
+from apiver_deps import SyncReport
 
-from .deps import SyncReport
 
-
-class TestSyncReport(TestBase):
+class TestSyncReport:
```
ppolewicz, line 18: it no longer inherits from TestBase!

b2sdk/v1/sync/report.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+# override to retain legacy methods
+class SyncReport(v2.SyncReport):
+    @property
+    def local_file_count(self):
+        return self.total_count
+
+    @local_file_count.setter
+    def local_file_count(self, value):
```
ppolewicz, line 21: does this really work? It looks like it would shadow the earlier one

b2sdk/bucket.py
```diff
@@ -135,7 +135,7 @@ def update(
         :param b2sdk.v1.BucketRetentionSetting default_retention: bucket default retention setting
         """
         account_id = self.api.account_info.get_account_id()
-        return BucketFactory.from_api_bucket_dict(
+        return self.api.BUCKET_FACTORY_CLASS.from_api_bucket_dict(
```
ppolewicz, line 138: ❤️

b2sdk/api_config.py
```diff
@@ -8,30 +8,33 @@
...
 ######################################################################
 
+from typing import Optional, Callable, Type
 import requests
 
-from typing import Optional
-
-from types import ModuleType
+if False:
+    from .raw_api import AbstractRawApi
```
ppolewicz: ?

b2sdk/session.py
```diff
@@ -67,7 +67,7 @@ def __init__(
         :param api_config
         """
 
-        self.raw_api = B2RawHTTPApi(B2Http(api_config))
+        self.raw_api = (api_config.raw_api_class or B2RawHTTPApi)(B2Http(api_config))
```
ppolewicz, line 70: I don't think we should be setting a default here. Config must have it.

b2sdk/encryption/setting.py
```diff
@@ -20,7 +21,12 @@
 
 logger = logging.getLogger(__name__)
 
-UNKNOWN_KEY_ID = ...
+
+class _UnknownKeyId(enum.Enum):
+    unknown_key_id = 0
+
+
+UNKNOWN_KEY_ID = _UnknownKeyId.unknown_key_id
```
ppolewicz, line 29: docstring please - what is it

b2sdk/encryption/setting.py
```diff
@@ -20,7 +21,12 @@
 
 logger = logging.getLogger(__name__)
 
-UNKNOWN_KEY_ID = ...
+
+class _UnknownKeyId(enum.Enum):
```
ppolewicz, line 25: docstring/comment please - why is it made this way?

b2sdk/api_config.py
mpnowacki-reef: The symbol is needed for type annotations but the module cannot be imported because that would cause circular dependencies

b2sdk/session.py
mpnowacki-reef: Yes. I didn't want to put it explicitly as a default value for `B2HttpApiConfig.__init__`, because `B2RawHttpApi` is part of the internal API, but it should be an evaluated default there. As some say: "there is a better way"

b2sdk/api_config.py
mpnowacki-reef: I've fixed the circular imports - they were caused by the tests in raw_api.

b2sdk/api_config.py
```diff
@@ -34,7 +33,7 @@ def __init__(
         self.http_session_factory = http_session_factory
         self.install_clock_skew_hook = install_clock_skew_hook
         self.user_agent_append = user_agent_append
-        self.raw_api_class = _raw_api_class
+        self.raw_api_class = _raw_api_class or B2RawHTTPApi
```
ppolewicz, line 36: I think it should be `self.DEFALT_RAW_API_CLASS` for future proofing like we did earlier in this PR

b2sdk/file_version.py
mpnowacki-reef: Parsing to datetime didn't work as it was implemented, and in general I've found it hard to parse these timestamps using builtin python libraries

test/unit/sync/test_sync_report.py
mpnowacki-reef: correct. It was moved to the new test structure, which uses pytest and no longer requires test classes to inherit from TestBase. In fact, it sometimes hurts when they do.

b2sdk/v1/sync/report.py
mpnowacki-reef: what do you mean? The getter and the setter? they don't work if their names are different

ppolewicz (approved):

PR #279, opened 2021-07-29
Disable decompression

mpnowacki-reef:


b2sdk/b2http_adapter.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+class NotDecompressingHTTPAdapter(HTTPAdapter):
+    """
+    HTTP adapter that uses :class:`b2sdk.requests.NotDecompressingResponse` instead of the default
+    :code:`requests.Response` class.
+    """
+
+    def build_response(self, req, resp):
+        return NotDecompressingResponse.from_builtin_response(super().build_response(req, resp))
```
ppolewicz, line 22: I think it should live in b2http so that all `requests`-related stuff is in one module

test/integration/test_download.py
```diff
@@ -79,3 +77,35 @@ def test_small_unverified(self):
...
+            temp_dir = pathlib.Path(temp_dir)
+            source_file = temp_dir / 'compressed_file.gz'
+            downloaded_uncompressed_file = temp_dir / 'downloaded_uncompressed_file'
+            downloaded_compressed_file = temp_dir / 'downloaded_compressed_file'
+
+            data_to_write = b"I'm about to be compressed and sent to the cloud, yay!\n" * 100  # too short files failed somehow
+            with gzip.open(source_file, 'wb') as gzip_file:
+                gzip_file.write(data_to_write)
+            file_version = bucket.upload_local_file(
+                str(source_file), 'gzipped_file', file_infos={'b2-content-encoding': 'gzip'}
```
ppolewicz, line 93: I think this was supposed to use `X-Bz-Info-b2-content-encoding` header, not file_info dict

ppolewicz (approved):

PR #297, opened 2022-01-21
Extract thread pool behavior and use in ParallelDownloader

awierzbicki-reef:
**TODO:**

- [x] Add comments, docstrings, and type annotations.
- [x] Make `set_thread_pool_size` and `get_thread_pool` protected members but keep `set_thread_pool_size` public in `_v3`.
- [x] Change default `max_workers` to `None` everywhere so it's adjusted to the number of CPU threads (for all pools: download, upload, copy), but keep the old default in `_v3`.
- [x] Change year from 2020 to 2022 in copyright message in `b2sdk/utils/thread_pool.py`


b2sdk/utils/thread_pool.py
```diff
@@ -0,0 +1,45 @@
...
+    def __init__(self, max_workers: 'Optional[int]' = None) -> None:
+        """
+        :param Optional[int] max_workers: maximum number of worker threads
+        """
+
+        self._lock = Lock()
+        self._thread_pool: 'Optional[ThreadPoolExecutor]' = None
+        self._max_workers = max_workers
+
+    def set_thread_pool_size(self, max_workers: int) -> None:
```
mpnowacki-reef, line 25: is this used anywhere? I think it's not. But if it is, then I think get_thread_pool should also use the same lock

b2sdk/utils/thread_pool.py
```diff
@@ -0,0 +1,45 @@
+from concurrent.futures import ThreadPoolExecutor
+from threading import Lock
+from typing import TYPE_CHECKING
+
+from b2sdk.utils import B2TraceMetaAbstract
+
+if TYPE_CHECKING:
```
mpnowacki-reef, line 7: I don't think this "if" is necessary

b2sdk/transfer/inbound/download_manager.py
```diff
@@ -42,18 +42,20 @@ class DownloadManager(metaclass=B2TraceMetaAbstract):
     MIN_CHUNK_SIZE = 8192  # ~1MB file will show ~1% progress increment
     MAX_CHUNK_SIZE = 1024**2
 
-    def __init__(self, services):
+    def __init__(self, services, max_download_workers=10):
```
ppolewicz, line 45: default is set to 10 in Services, in B2Api and in DownloadManager. Is this right?

b2sdk/transfer/inbound/download_manager.py
```diff
@@ -42,18 +42,20 @@ class DownloadManager(metaclass=B2TraceMetaAbstract):
...
         :param b2sdk.v2.Services services:
+        :param int max_download_workers: maximum number of download threads
         """
 
         self.services = services
         self.strategies = [
             ParallelDownloader(
                 max_streams=self.DEFAULT_MAX_STREAMS,
                 min_part_size=self.DEFAULT_MIN_PART_SIZE,
+                max_download_workers=max_download_workers,
```
ppolewicz, line 58: put it at the end

b2sdk/transfer/inbound/downloader/parallel.py
```diff
@@ -49,14 +51,23 @@ class ParallelDownloader(AbstractDownloader):
...
+            *args,
+            **kwargs,
+    ) -> None:
         """
         :param max_streams: maximum number of simultaneous streams
         :param min_part_size: minimum amount of data a single stream will retrieve, in bytes
+        :param int max_download_workers: maximum number of download threads
         """
+        super().__init__(*args, **kwargs)
+        LazyThreadPool.__init__(self, max_workers=max_download_workers)
```
ppolewicz, line 68: I think this is weird, super() calls should chain each other, should not need you to call two explicitly

b2sdk/transfer/inbound/downloader/parallel.py
```diff
@@ -243,96 +254,100 @@ def run(self):
...
-
-
-class NonHashingDownloaderThread(AbstractDownloaderThread):
-    def __init__(self, url, *args, **kwargs):
-        """
-        :param url: url of the target file
-        """
-        self.url = url
-        super(NonHashingDownloaderThread, self).__init__(*args, **kwargs)
+# TODO: Add param type annotations & docstrings
```
ppolewicz, line 257: marking TODO

b2sdk/transfer/inbound/downloader/parallel.py
```diff
@@ -243,96 +254,100 @@ def run(self):
...
+            encryption=encryption,
+        ) as response:
+            for to_write in response.iter_content(chunk_size=chunk_size):
+                writer_queue_put((False, first_offset + bytes_read, to_write))
+                hasher_update(to_write)
+                bytes_read += len(to_write)
+        tries_left -= 1
+
+
+# TODO: Add param type annotations & docstrings
```
ppolewicz, line 319: marking TODO

b2sdk/transfer/outbound/copy_manager.py
```diff
@@ -32,38 +31,15 @@ class CopyManager(metaclass=B2TraceMetaAbstract):
     def __init__(self, services, max_copy_workers=10):
```
ppolewicz, line 31: do we need the parameters in all Manager classes have different names for the worker count? Maybe it could all be `max_workers`, then maybe more of `__init__` could go up in class hierarchy?

b2sdk/utils/thread_pool.py
ppolewicz: agreed, `get` needs to use the same lock

b2sdk/utils/thread_pool.py
```diff
@@ -0,0 +1,45 @@
...
+    from typing import Optional
+
+
+class LazyThreadPool(metaclass=B2TraceMetaAbstract):
+    """
+    TODO: Add docstring
```
ppolewicz, line 13: marking TODO

ppolewicz (approved):
approved to run tests

b2sdk/transfer/inbound/download_manager.py
awierzbicki-reef: Yes, I wanted to be consistent with `max_upload_workers` and `max_copy_workers`. In general, I believe the default should be rather `None` (let the ThreadPool implementation choose appropriate default), but I'm not sure if I can change this without breaking public API compatibility.

b2sdk/transfer/inbound/downloader/parallel.py
awierzbicki-reef: I believe with multiple inheritance, `super()` only calls the first parent class' constructor.

b2sdk/transfer/outbound/copy_manager.py
awierzbicki-reef: :thinking: May be a good idea.

b2sdk/transfer/inbound/download_manager.py
ppolewicz: lets not keep consistency with our past bad practices

ppolewicz: those classes are not members of the public interface so you can change them at will

b2sdk/transfer/inbound/downloader/parallel.py
ppolewicz: yes, but that parent should call it parent, and that one should call its own parent etc, all the way to `object`

b2sdk/transfer/outbound/copy_manager.py
ppolewicz: if we are refactoring, lets refactor

b2sdk/transfer/inbound/downloader/parallel.py
awierzbicki-reef: If suspect you're missing the point there's multiple inheritance here. `LazyThreadPoolMixin` is **not** a parent of  `AbstractDownloader` so where is its `__init__` supposed to be called?

ppolewicz: please watch https://www.youtube.com/watch?v=EiOglTERPEo (billing `RT / onboarding`) and then lets talk again

b2sdk/api.py
```diff
@@ -51,21 +51,27 @@ def url_for_api(info, api_name):
...
         """
         self.api = api
         self.session = api.session
         self.large_file = LargeFileServices(self)
-        self.download_manager = DownloadManager(self, max_download_workers=max_download_workers)
-        self.upload_manager = UploadManager(self, max_upload_workers=max_upload_workers)
-        self.copy_manager = CopyManager(self, max_copy_workers=max_copy_workers)
+        self.download_manager = DownloadManager(self, max_workers=max_download_workers)
+        self.upload_manager = UploadManager(self, max_workers=max_upload_workers)
+        self.copy_manager = CopyManager(self, max_workers=max_download_workers)
```
ppolewicz, line 74: the order in the signature should be the same as the order in here

b2sdk/transfer/outbound/copy_manager.py
```diff
@@ -28,12 +28,12 @@ class CopyManager(LazyThreadPoolMixin):
...
-    def __init__(self, services, max_copy_workers=10):
+    def __init__(self, services, max_workers: Optional[int] = None):
         """
         :param b2sdk.v2.Services services:
-        :param int max_copy_workers: maximum number of copy threads
+        :param int max_workers: maximum number of copy threads
         """
-        super().__init__(max_workers=max_copy_workers)
+        super().__init__(max_workers=max_workers)
         self.services = services
```
ppolewicz, line 37: this should be factored out to a parent class or something

b2sdk/_v3/api.py
```diff
@@ -0,0 +1,25 @@
...
+from .transfer import UploadManager
+
+
+class B2Api(api.B2Api):
+
+    # Legacy init in case something depends on max_workers defaults = 10
+    def __init__(self, *args, **kwargs):
+        kwargs.setdefault('max_upload_workers', 10)
+        kwargs.setdefault('max_copy_workers', 10)
+        with mock.patch('b2sdk.api.UploadManager', UploadManager):
```
ppolewicz, line 24: this is not the way. You have to swap out UploadManager without using mock.patch - see how other apiver wrappers are doing it

b2sdk/transfer/outbound/copy_manager.py
awierzbicki-reef: You mean `self.services`? Sure, why not.

b2sdk/_v3/api.py
awierzbicki-reef: You mean by introducing some kind `_FACTORY_CLASS` class variable? Yeah, that's cleaner. I did it the easy/lazy way here.

b2sdk/transfer/inbound/downloader/parallel.py
awierzbicki-reef: :tv: :eyes:

awierzbicki-reef: Thanks for sharing this. I've learned quite a bit. I'll change this to avoid calling `LazyThreadPool.__init__` explicitly.

awierzbicki-reef:
@ppolewicz @mpnowacki-reef Would you find some time for the final (I hope so) review?

b2sdk/utils/thread_pool.py
```diff
@@ -0,0 +1,55 @@
...
+        """
+        :param Optional[int] max_workers: maximum number of worker threads
+        """
+
+        self._lock = Lock()
+        self._thread_pool = None  # type: 'Optional[ThreadPoolExecutor]'
+        self._max_workers = max_workers
+        super().__init__(**kwargs)
+
+    def _set_thread_pool_size(self, max_workers: int) -> None:
```
mpnowacki-reef, line 34: what's the point of this method? Is it here only for the sake of providing a public version of it in spicer wrappers?

b2sdk/_v3/transfer.py
```diff
@@ -0,0 +1,36 @@
...
+#
+######################################################################
+
+from b2sdk import transfer
+from b2sdk.transfer.inbound.downloader import parallel
+
+
+class ParallelDownloader(parallel.ParallelDownloader):
+
+    # This method is used in SDK even though it doesn't belong to the public API
```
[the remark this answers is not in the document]
awierzbicki-reef: This is what I meant.

b2sdk/utils/thread_pool.py
awierzbicki-reef: Yes, we agreed with @ppolewicz  that it is bad design to expose it, but for compatibility reasons we decided to keep it as protected.

b2sdk/utils/thread_pool.py
```diff
@@ -0,0 +1,55 @@
...
+    Can be safely used between threads.
+    """
+
+    def __init__(self, max_workers: 'Optional[int]' = None, **kwargs) -> None:
+        """
+        :param Optional[int] max_workers: maximum number of worker threads
+        """
+
+        self._lock = Lock()
+        self._thread_pool = None  # type: 'Optional[ThreadPoolExecutor]'
```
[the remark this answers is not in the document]
awierzbicki-reef: This doesn't work in Python 3.5. :disappointed:

b2sdk/_v3/api.py
```diff
@@ -0,0 +1,28 @@
...
+# License https://www.backblaze.com/using_b2_code.html
+#
+######################################################################
+
+from b2sdk import api
+
+from .transfer import DownloadManager, UploadManager
+
+
+class Services(api.Services):
```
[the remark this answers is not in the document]
awierzbicki-reef: Fixed.

b2sdk/utils/thread_pool.py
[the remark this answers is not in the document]
mpnowacki-reef: ah snap

b2sdk/v2/transfer.py
```diff
@@ -0,0 +1,35 @@
...
+
+    # This method is used in CLI even though it doesn't belong to the public API
+    def set_thread_pool_size(self, max_workers: int) -> None:
+        self._set_thread_pool_size(max_workers)
+
+
+class DownloadManager(v3.DownloadManager):
+    PARALLEL_DOWNLOADER_CLASS = staticmethod(ParallelDownloader)
+
+    # This method is used in CLI even though it doesn't belong to the public API
```
mpnowacki-reef, line 24: Okay, I thought the current plan is to include this class/method in the public API and remove it from public API in v3. Otherwise it's bound to break at some point.

b2sdk/v2/transfer.py
```diff
@@ -0,0 +1,35 @@
...
+    def set_thread_pool_size(self, max_workers: int) -> None:
+        for strategy in self.strategies:
+            if isinstance(strategy, ParallelDownloader):
+                strategy.set_thread_pool_size(max_workers)
+
+
+class UploadManager(v3.UploadManager):
+
+    # This method is used in SDK even though it doesn't belong to the public API
```
mpnowacki-reef, line 33: Okay, I thought the current plan is to include this class/method in the public API and remove it from public API in v3. Otherwise it's bound to break at some point.

b2sdk/v2/transfer.py
```diff
@@ -0,0 +1,35 @@
...
+# License https://www.backblaze.com/using_b2_code.html
+#
+######################################################################
+
+from b2sdk import _v3 as v3
+
+
+class ParallelDownloader(v3.ParallelDownloader):
+
+    # This method is used in CLI even though it doesn't belong to the public API
```
mpnowacki-reef, line 16: Okay, I thought the current plan is to include this class/method in the public API and remove it from public API in v3. Otherwise it's bound to break at some point.

b2sdk/utils/thread_pool.py
mpnowacki-reef: in that case, I don't think it should be here. If it's only purpose is to be used in spicer wrappers, then it should be defined there

b2sdk/v2/transfer.py
```diff
@@ -0,0 +1,35 @@
+######################################################################
```
mpnowacki-reef, line 1: the `set_thread_pool_size` methods are now part of the public API, but are going to be removed in v3. They are not tested, not even for existence and signature, which means CLI is probably gonna break when we release v3

mpnowacki-reef (dismissed):

b2sdk/v2/transfer.py
awierzbicki-reef: They're not part of the public API in the sense that their not mentioned in the docs as such. Is CLI gonna break when they're removed? I think it is not. I assume CLI is importing only from `v2`, and we're keeping this methods in `v2`, right?

b2sdk/v2/transfer.py
awierzbicki-reef: The plan is to treat this as though it were a part of the public API, even though it actually isn't. So we're acknowledging that there is some external code using it and we don't want that code to break. But once we release `v3` and people start using it, we want to force them to stop using it.

b2sdk/v2/transfer.py
ppolewicz: it is used in CLI, not sdk

CHANGELOG.md
```diff
@@ -6,12 +6,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+### Changed
+* Introduced a thread worker pool for parallel downloads
+
 ### Added
 * Add support for unverified checksum upload mode
 
 ### Fixed
 * Fix downloading files with unverified checksum
 
+
```
ppolewicz, line 18: This empty line should not be here, but the changelog linter should have told you that. I think it's only run in CI while it should also be run in `nox`. Please file a separate issue (or PR) for it.

b2sdk/v2/transfer.py
mpnowacki-reef: as it is now, the apiver methods use a deprecated private method defined in the "upstream" managers. If a developer simply removes it, these methods will stop working in v2, no test will break in the SDK and CLI will raise errors when using these options.

b2sdk/transfer/inbound/download_manager.py
```diff
@@ -42,22 +43,26 @@ class DownloadManager(metaclass=B2TraceMetaAbstract):
     MIN_CHUNK_SIZE = 8192  # ~1MB file will show ~1% progress increment
     MAX_CHUNK_SIZE = 1024**2
 
-    def __init__(self, services):
+    PARALLEL_DOWNLOADER_CLASS = staticmethod(ParallelDownloader)
+    SIMPLE_DOWNLOADER_CLASS = staticmethod(SimpleDownloader)
+
+    def __init__(self, max_workers: Optional[int] = None, **kwargs) -> None:
```
ppolewicz, line 49: please don't specify the return type from `__init__`

b2sdk/transfer/inbound/downloader/parallel.py
```diff
@@ -49,14 +50,14 @@ class ParallelDownloader(AbstractDownloader):
     #
     FINISH_HASHING_BUFFER_SIZE = 1024**2
 
-    def __init__(self, max_streams, min_part_size, *args, **kwargs):
+    def __init__(self, max_streams: int, min_part_size: int, *args, **kwargs) -> None:
```
ppolewicz, line 53: please don't specify the return type from `__init__`

b2sdk/transfer/transfer_manager.py
```diff
@@ -0,0 +1,19 @@
...
+#
+######################################################################
+
+
+class TransferManager:
+    """
+    Base class for manager classes (copy, upload, download)
+    """
+
+    def __init__(self, services: 'b2sdk.api.Services', **kwargs) -> None:
```
ppolewicz, line 17: please don't specify the return type from `__init__`

b2sdk/utils/thread_pool.py
```diff
@@ -0,0 +1,55 @@
...
+from b2sdk.utils import B2TraceMetaAbstract
+
+
+class LazyThreadPoolMixin(metaclass=B2TraceMetaAbstract):
+    """
+    Mixing class with lazy initialized ThreadPoolExecutor.
+    Can be safely used between threads.
+    """
+
+    def __init__(self, max_workers: 'Optional[int]' = None, **kwargs) -> None:
```
ppolewicz, line 24: please don't specify the return type from `__init__`

b2sdk/utils/thread_pool.py
```diff
@@ -0,0 +1,55 @@
...
+from concurrent.futures import ThreadPoolExecutor
+from threading import Lock
+from typing import Optional
+
+from b2sdk.utils import B2TraceMetaAbstract
+
+
+class LazyThreadPoolMixin(metaclass=B2TraceMetaAbstract):
+    """
+    Mixing class with lazy initialized ThreadPoolExecutor.
```
ppolewicz, line 20:
```suggestion
    Mixin class with lazily initialized ThreadPoolExecutor.
```

b2sdk/utils/thread_pool.py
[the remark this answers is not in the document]
ppolewicz: 3.5 is standing in the doorway, just needs a little bit of a push...

b2sdk/utils/thread_pool.py
ppolewicz: yes, those should be moved to `b2sdk/v2`

b2sdk/utils/thread_pool.py
```diff
@@ -0,0 +1,55 @@
...
+        """
+        with self._lock:
+            if self._thread_pool is not None:
+                raise RuntimeError('Thread pool already created')
+            self._max_workers = max_workers
+
+    def _get_thread_pool(self) -> ThreadPoolExecutor:
+        """
+        Return the thread pool executor.
+        """
```
ppolewicz, line 51: I don't think this docstring provides any value, please remove it

CHANGELOG.md
```diff
@@ -6,12 +6,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+* Introduced a thread worker pool for parallel downloads
```
ppolewicz, line 10:
```suggestion
* Introduce a common thread worker pool for all parallel downloads
```

ppolewicz: If we have sync threads set to 10 and download threads set to 10, how many threads can be used to download things at the same time? 10? No. 19, because N sync threads can occupy the thread pool while M sync threads can just do small and easy single-threaded downloads and while N+M=10, you can also have N=1,M=9 so 10 large file downloads and 9 small file downloads at the same time. This is not what the user wants though. I think small downloads should also be put in that same download threadpool.

b2sdk/v2/transfer.py
ppolewicz: It needs to be tested because when we release b2sdk v3, we might accidentally break something in the v2 interface and then b2cli installation which still uses b2sdk v2 (but a fresh v3-capable version of it) may break and since there are no tests for it at all, we won't know before the release.

ppolewicz (changes requested):

b2sdk/v2/transfer.py
awierzbicki-reef: Alright, I'll add some unit test.

b2sdk/utils/thread_pool.py
awierzbicki-reef: Already done.

CHANGELOG.md
awierzbicki-reef: Done in 1feddc9cf8802e72fba55bb0b705cced5e9c679d

b2sdk/v2/transfer.py
awierzbicki-reef: Done in [b4c3758](https://github.com/Backblaze/b2-sdk-python/pull/297/commits/b4c3758f811af385a9a098ff281e03e6f67fc569)

b2sdk/transfer/inbound/downloader/simple.py
```diff
@@ -35,45 +35,49 @@ def download(
         session: B2Session,
         encryption: Optional[EncryptionSetting] = None,
     ):
-        actual_size = self._get_remote_range(response, download_version).size()
-        chunk_size = self._get_chunk_size(actual_size)
+        def download():
```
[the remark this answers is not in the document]
awierzbicki-reef: I think it can still be tested by some dependency injection, but extracting it is OK.

b2sdk/transfer/inbound/downloader/simple.py
```diff
@@ -7,9 +7,9 @@
 # License https://www.backblaze.com/using_b2_code.html
 #
 ######################################################################
-
+from concurrent.futures import ThreadPoolExecutor
```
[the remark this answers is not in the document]
awierzbicki-reef: First of all, this line should be deleted because the import is not used. :stuck_out_tongue:

b2sdk/transfer/inbound/downloader/simple.py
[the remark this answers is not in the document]
ppolewicz: Maybe you could swap it with a mock or something, but if you'd ever need to modify the behavior in some instances of the class, you'd like to be able to easily swap the download helper method, and here you have to swap it with the caller. + testability also requires more thinking

b2sdk/utils/thread_pool.py
```diff
@@ -8,38 +8,17 @@
...
 
 from b2sdk.utils import B2TraceMetaAbstract
 
 
-class LazyThreadPool(metaclass=B2TraceMetaAbstract):
+class ThreadPoolMixin(metaclass=B2TraceMetaAbstract):
     """
-    Lazily initialized thread pool.
-    Can be safely used between threads.
+    Mixin class with lazily initialized ThreadPoolExecutor. Threadsafe.
```
ppolewicz, line 19: is this one lazy though?

b2sdk/v2/transfer.py
```diff
@@ -7,29 +7,48 @@
 # License https://www.backblaze.com/using_b2_code.html
 #
 ######################################################################
+from concurrent.futures import Future, ThreadPoolExecutor
```
ppolewicz, line 10: should be a blank line above that

b2sdk/v2/transfer.py
```diff
@@ -7,29 +7,48 @@
...
+from threading import Lock
+from typing import Callable
 
 from b2sdk import _v3 as v3
 
 
-class LazyThreadPool(v3.LazyThreadPool):
+class LazyThreadPool:
```
ppolewicz, line 17: it looks like it shares the interface with `ThreadPoolMixin`, but not explicitly?

b2sdk/utils/thread_pool.py
awierzbicki-reef: Nope, but it was before the last refactor.

b2sdk/v2/transfer.py
awierzbicki-reef: My IDE automatically adds imports and `nox -s lint` doesn't check it... :/

b2sdk/utils/thread_pool.py
awierzbicki-reef: Now it's just a mixin with a thread pool.

awierzbicki-reef: I think I deleted too much, it's no longer threadsafe as well. :( Will fix.

awierzbicki-reef: Fixed in [fc9de68](https://github.com/Backblaze/b2-sdk-python/pull/297/commits/fc9de68873550ac431cee0d257bb714d1ebee152)

b2sdk/v2/transfer.py
awierzbicki-reef: Fixed in [fc9de68](https://github.com/Backblaze/b2-sdk-python/pull/297/commits/fc9de68873550ac431cee0d257bb714d1ebee152)

b2sdk/v2/transfer.py
awierzbicki-reef: After [fc9de68](https://github.com/Backblaze/b2-sdk-python/pull/297/commits/fc9de68873550ac431cee0d257bb714d1ebee152) it shares the `submit` method. Is that a problem?

awierzbicki-reef: Not anymore.

b2sdk/transfer/inbound/downloader/parallel.py
```diff
@@ -66,7 +66,14 @@ def is_suitable(self, download_version: DownloadVersion, allow_seeking: bool):
...
 
     def _get_number_of_streams(self, content_length):
-        return min(self.max_streams, content_length // self.min_part_size) or 1
+        num_streams = content_length // self.min_part_size
+        if self.max_streams is not None:
+            num_streams = min(num_streams, self.max_streams)
+        else:
+            max_threadpool_workers = getattr(self._thread_pool, '_max_workers', None)
+            if max_threadpool_workers is not None:
+                num_streams = min(num_streams, max_threadpool_workers)
```
ppolewicz, line 75: if my understanding is correct, limiting the number of workers to 1 and uploading a file larger than 5GB will cause a crash

ppolewicz (approved):
approving to run tests

b2sdk/transfer/inbound/downloader/parallel.py
mzukowski-reef: this is downloading :)

ppolewicz: ohhhhhh 🤦

PR #299, opened 2022-01-31
Fix downloading objects with unverified checksum

ppolewicz:


b2sdk/v1/file_metadata.py
```diff
@@ -11,7 +11,7 @@
 import b2sdk.v2 as v2
 
 
-class FileMetadata(object):
+class FileMetadata:
```
mpnowacki-reef, line 14: :+1:

test/unit/bucket/test_bucket.py
```diff
@@ -1646,6 +1646,25 @@ def test_download_by_name_empty_file(self):
         self._verify('')
 
 
+class UnverifiedChecksumDownloadScenarioMixin(object):
```
mpnowacki-reef, line 1649:
```suggestion
class UnverifiedChecksumDownloadScenarioMixin:
```

mpnowacki-reef:
This functionality deserves an integration test, to make sure The API does not change the syntax slightly. How complex is it to replicate an unverified checksum?

PR #301, opened 2022-02-05
Add dedicated exception for unverified email

ppolewicz:


mpnowacki-reef:
no integration or raw-simulator-based tests?

mpnowacki-reef:
or are you planning on implementing those in CLI?

ppolewicz:
The code is super simple but very hard to test in integration as you'd need
a separate account with unconfirmed email. Tested it manually. Do we really
need coverage for this?

PR #303, opened 2022-02-19
Fix filename in get_file_info_by_name

KernelErr:
During using b2 object storage API, I found that file names such as `test.txt?a=1` could be uploaded. However, the following code would throw a `FileNotPresent` error as the file name returned in the HTTP header is its encoded `test.txt%3Fa=1` which could be another file.

```python
file_info = bucket.get_file_info_by_name('test.txt?a=1')
print(file_info)
bucket.delete_file_version(file_info.id_, file_info.file_name)
```

In this PR, I manually set the file name field in `DownloadVersion` to the file name in the parameter to fix this problem.

Signed-off-by: KernelErr <me@lirui.tech>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 * Fix downloading files with unverified checksum
+* Fix decoding in filename and file info of `DownloadVersion`
 
 ### Changed
 * Don't run coverage in pypy in CI
--- a/b2sdk/bucket.py
+++ b/b2sdk/bucket.py
@@ -236,9 +236,11 @@ class Bucket(metaclass=B2TraceMeta):
         :param str file_name: the name of the file who's info will be retrieved.
         """
         try:
-            return self.api.download_version_factory.from_response_headers(
+            download_version = self.api.download_version_factory.from_response_headers(
                 self.api.session.get_file_info_by_name(self.name, file_name)
             )
+            download_version._set_filename(file_name)
+            return download_version
         except FileOrBucketNotFound:
             raise FileNotPresent(bucket_name=self.name, file_id_or_name=file_name)
 
--- a/b2sdk/bucket.py
+++ b/b2sdk/bucket.py
@@ -236,11 +236,9 @@ class Bucket(metaclass=B2TraceMeta):
         :param str file_name: the name of the file who's info will be retrieved.
         """
         try:
-            download_version = self.api.download_version_factory.from_response_headers(
+            return self.api.download_version_factory.from_response_headers(
                 self.api.session.get_file_info_by_name(self.name, file_name)
             )
-            download_version._set_filename(file_name)
-            return download_version
         except FileOrBucketNotFound:
             raise FileNotPresent(bucket_name=self.name, file_id_or_name=file_name)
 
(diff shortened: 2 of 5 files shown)


ppolewicz:
🤔 is `get_file_info_by_id` not affected by the same phenomenon? I think we should decode the filename properly in `FileVersion` factory instead of setting it to the name if we happen to know it. Then it will work for both `get_file_info_by_*` methods (and for all future methods, I guess)

KernelErr:
> 🤔 is `get_file_info_by_id` not affected by the same phenomenon? I think we should decode the filename properly in `FileVersion` factory instead of setting it to the name if we happen to know it. Then it will work for both `get_file_info_by_*` methods (and for all future methods, I guess)

We may not be able to decode every file name. As `test.txt%3Fa=1` and `test.txt?a=1` may coexist.

ppolewicz:
I've been looking at this PR several times now, unable to decide whether it's the correct approach or not and today I realized why I can't decide. It lacks quality assurance.

In order to make sure this fully solves the problem, things that should be added are:
- a unit test (using simulator)
- an integration test that verifies the behavior is correct with the server, also verifying that the correct file is downloaded if both encoded and non-encoded versions coexist
- generally following the guidance of https://www.backblaze.com/b2/docs/string_encoding.html to make sure we are doing the right thing where we should and that we are not doing any extra encoding

Additionally I've noticed that your PR is calling a _protected method from outside of it's class (which is not causing compilation errors in Python, but still it's a bad idea) and that the class itself is now required to have this method, but it is not set as an abstract in the interface, so if people implement their versions of this class and plug them in, they won't know it doesn't work until they try to use the get_file_info method and see that it crashes. If the method was set to be abstract, creating an instance of a class without required methods would fail... But the method itself is a setter and in Python we generally don't need setters, you can just do obj.field = value and it works (though objects of most classes should not be modified like that). In this case the place where this snippet of code is added may not be the best one, perhaps it's the factory method which lacks decoding.

KernelErr:
Thanks for your comment. Sadly I have little experience in Python projects like this scale. I changed `DownloadVersionFactory`, now it will decode the filename got from Backblaze API as "The values of the custom headers X-Bz-File-Name and X-Bz-Info-* must be percent-encoded".

ppolewicz:
Then I think `file_info[file_info_key] = header_value` in line 455 should also be changed to apply the decoding, right?

Please adjust the changelog so that it mentiones the fix, such as:
```
### Fixed
* Fix decoding filename and file info values
```

ppolewicz:
Ok to get this merged we just need one more thing: tests to make sure this behavior is correct and stays correct. We wouldn't want your patch to be reverted accidentally during a refactor, right?

Can you look into our test suites and try to add an appropriate test for your fix? The link to the b2 website seems to list the test cases for encoding nicely.

Meanwhile I'll fix the changelog, it looks like it suffered from a bad merge.,

KernelErr:
Hi, I added decode tests in the simulator. In the meanwhile, I fixed a encode problem that the simulator wound not encode file information. Could you help me fix the changelog? @ppolewicz

ppolewicz:
It's fixed on master now, you can merge master to your branch and that should fix the changelog. Let me know when you do that, I'll approve the test run and then we merge it

KernelErr:
Fixed. :) @ppolewicz

ppolewicz (approved):

PR #311, opened 2022-03-31
Remove inheritance from object (leftover from python2)

ppolewicz:


agoncharov-reef (approved):

agoncharov-reef:
You may also remove `def interruptible_get_result(future):`  :)

PR #312, opened 2022-03-31
Drop support for Python 3.5 and Python 3.6

ppolewicz:


b2sdk/sync/scan_policies.py
```diff
@@ -18,10 +18,7 @@
 
 logger = logging.getLogger(__name__)
 
-try:  # python 3.5 and 3.6 compatibility
-    regex_class = re.Pattern
-except AttributeError:
-    regex_class = re._pattern_type
+regex_class = re.Pattern
```
agoncharov-reef, line 21: Why not removing `regex_class` and using it directly `re.Pattern` in appropriate place? The variable is redundant now

b2sdk/v1/sync/scan_policies.py
```diff
@@ -17,10 +17,7 @@
 from b2sdk import v2
 from b2sdk.v2 import exception as v2_exception  # noqa
 
-try:  # python 3.5 and 3.6 compatibility
-    regex_class = re.Pattern
-except AttributeError:
-    regex_class = re._pattern_type
+regex_class = re.Pattern
```
agoncharov-reef, line 20: Same here

setup.py
```diff
@@ -111,7 +109,7 @@ def read_requirements(extra=None):
     # for example:
     # $ pip install -e .[dev,test]
     extras_require={'doc': read_requirements('doc')},
-    setup_requires=['setuptools_scm<6.0'],  # setuptools_scm>=6.0 doesn't support Python 3.5
+    setup_requires=['setuptools_scm'],
```
agoncharov-reef, line 112: Why not pinning the version?

ppolewicz: I'd think we want to use the newest one we can find during install and if it crashes, investigate and maybe pin to a lower version then

ppolewicz: setuptools_scm removed support for setup.py sometimes after 6.0

PR #326, opened 2022-06-01
Enable pip cache in CI

ppolewicz:
installing dependencies on Windows takes up to 3 minutes...

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,9 +21,10 @@ jobs:
         with:
           ignore_words_list: datas
       - name: Set up Python ${{ env.PYTHON_DEFAULT_VERSION }}
-        uses: actions/setup-python@v2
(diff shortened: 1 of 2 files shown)


ppolewicz:
https://github.com/Backblaze/b2-sdk-python/runs/6695671537?check_suite_focus=true

ppolewicz:
Time reduced from ~3min to 20-30s, it seems that it's safe to use in this context since pip still touches the network for the package list and the indexes are specific to python version and OS. Won't it refuse to install B2 this way, showing green all the time?

mpnowacki-reef:
@ppolewicz Looks good to me. 

> Won't it refuse to install B2 this way, showing green all the time?

Why would b2-sdk-pythons CI install B2?

mpnowacki-reef (approved):

PR #329, opened 2022-06-19
Adjust timeouts to prevent failures on 1TB+ objects

ppolewicz:
Resolves Backblaze/B2_Command_Line_Tool#554

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,8 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 * Change the per part retry limit from 5 to 20 for data transfer operations. Please note that the retry system is not considered to be a part of the public interface and is subject to be adjusted
-* After the 10th retry do not wait more than 64 seconds between retry attempts
-* After the 5th retry wait a random amount of time (up to 1s) to prevent client synchronization
+* Do not wait more than 64 seconds between retry attempts (unless server asks for it)
+* On longer failures wait an additional (random, up to 1s) amount of time to prevent client synchronization
 
 ### Fixed
 * Fix `AccountInfo.is_master_key()`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 * Add `include_existing_files` parameter to `ReplicationSetupHelper`
 
+### Changed
+* Change the per part retry limit from 5 to 20 for data transfer operations. Please note that the retry system is not considered to be a part of the public interface and is subject to be adjusted
+* After the 10th retry do not wait more than 64 seconds between retry attempts
+* After the 5th retry wait a random amount of time (up to 1s) to prevent client synchronization
+
 ### Fixed
 * Fix `AccountInfo.is_master_key()`
 * Fix docstring of `SqliteAccountInfo`
--- a/b2sdk/b2http.py
+++ b/b2sdk/b2http.py
@@ -8,7 +8,7 @@
 #
 ######################################################################
 
-from random import Random
+from random import random
 import io
 import json
 import logging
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 * Add `include_existing_files` parameter to `ReplicationSetupHelper`
 
+### Changed
+* Change the per part retry limit from 5 to 20 for data transfer operations. Please note that the retry system is not considered to be a part of the public interface and is subject to be adjusted
+* After the 10th retry do not wait more than 64 seconds between retry attempts
+* After the 5th retry wait a random amount of time (up to 1s) to prevent client synchronization
```
mpnowacki-reef, line 15: Did you mean `After the 5th retry increase the wait time by a random amount of time (up to 1s)` ?

Either way, I don't see it, I can only pinpoint where you add the random number when hitting max

b2sdk/b2http.py
```diff
@@ -241,10 +255,24 @@ def post_json_return_json(self, url, headers, params, try_count=5):
...
+            url,
+            headers,
+            data,
+            try_count,
+            params,
+            force_timeout=timeout,
+        )
 
-    def get_content(self, url, headers, try_count=5):
+    def get_content(self, url, headers, try_count=20):
```
mpnowacki-reef, line 275: I know it used to be like that, but I really don't like these literals just pasted here. Don't you think it would be better to use them as constants?

b2sdk/b2http.py
```diff
@@ -413,6 +441,8 @@ def _translate_and_retry(cls, fcn, try_count, post_params=None):
         """
         # For all but the last try, catch the exception.
         wait_time = 1.0
+        max_wait_time = 64
```
mpnowacki-reef, line 444: same here

b2sdk/b2http.py
```diff
@@ -176,7 +179,15 @@ def add_callback(self, callback):
...
 
-    def post_content_return_json(self, url, headers, data, try_count=5, post_params=None):
+    def post_content_return_json(
+        self,
+        url,
+        headers,
+        data,
+        try_count=20,
+        post_params=None,
+        force_timeout=None,
```
mpnowacki-reef, line 189: that's a bit awkward, I think this parameter should be named `timeout` and have it's default value computed inside of the function just like it is now.

Anyway, typing would improve the current nomenclature. `force_timeout`, without even being mentioned in the docstring should be typed, someone could use a `bool` as its value as it is now.

b2sdk/b2http.py
```diff
@@ -413,6 +441,8 @@ def _translate_and_retry(cls, fcn, try_count, post_params=None):
         """
         # For all but the last try, catch the exception.
         wait_time = 1.0
+        max_wait_time = 64
+        rng = Random()
```
mpnowacki-reef, line 445: any particular reason for having a separately seeded generator for each call? The execution time of initialisation is humungous compared to a single number generation:

```
22:46:39 reef@~
$ python3 -m timeit 'import random; random.Random()'
10000 loops, best of 5: 25.1 usec per loop
22:47:01 reef@~
$ python3 -m timeit 'import random; random.Random().random()'
10000 loops, best of 5: 25.6 usec per loop
22:47:10 reef@~
$ python3 -m timeit 'import random; random.random()'
2000000 loops, best of 5: 123 nsec per loop
```

but completely negligible compared to HTTP IO. I'm just curious as to why.

mpnowacki-reef (changes requested):

b2sdk/b2http.py
ppolewicz: if we use constants, people will start customizing it and we really don't want it to happen - the entire retry system is up for a rewrite

b2sdk/b2http.py
ppolewicz: you are right, I didn't know that's so slow!

b2sdk/b2http.py
ppolewicz: _timeout + type hint

b2sdk/b2http.py
mpnowacki-reef: if it's up for a rewrite then yeah, let's leave those literals as they are

mpnowacki-reef (approved):

PR #336, opened 2022-06-23
Fix integration tests on non-production environments

ppolewicz:


test/integration/test_download.py
```diff
@@ -41,42 +43,47 @@ def test_large_file(self):
...
+
                 f = self._file_helper(bucket)
-                assert f.download_version.content_sha1_verified
+                if zero._type() != 'large':
+                    # if we are here, that's not the production server!
+                    assert f.download_version.content_sha1_verified  # large files don't have sha1, lets not check
 
-    def _file_helper(self, bucket, sha1_sum=None):
-        bytes_to_write = int(self.info.get_absolute_minimum_part_size()) * 2 + 1
+    def _file_helper(self, bucket, sha1_sum=None, bytes_to_write: Optional[int] = None) -> DownloadVersion:
```
ppolewicz, line 57:
```suggestion
    def _file_helper(
        self, bucket, sha1_sum=None, bytes_to_write: Optional[int] = None
    ) -> DownloadVersion:
```

PR #337, opened 2022-06-23
Add import testing on the source package after it's built

ppolewicz:


ppolewicz:
Fails when run locally but passes in CI. @mpnowacki-reef any clues?

noxfile.py
```diff
@@ -190,6 +191,14 @@ def build(session):
             version = os.environ['GITHUB_REF'].replace('refs/tags/v', '')
             print(f'version={version}', file=github_output)
 
+    sdists = sorted(pathlib.Path('dist').glob('b2sdk-*.tar.gz'))
+    if len(sdists) != 1:
+        session.error(f'Expected exactly one source distribution, found {len(sdists)}: {sdists!r}')
+
+    session.install(str(sdists[0]))
+    session.cd('dist')  # avoid importing from the checkout instead of the built package
+    session.run('python', '-c', 'from b2sdk import v0, v1, v2, v3')
```
mpnowacki-reef: I think this might be a bit fragile, depending on what's in PYTHONPATH etc. maybe print `v0.__file__` etc. and assert it is imported from the build directory?

PR #341, opened 2022-08-04
Prepare release v1.17.4

ppolewicz:


ppolewicz:
This could not have been merged due to CI issues which were now resolved, but it does not make sense to release it as is as we'll also be merging a couple of other things this week. One valuable commit was cherry-picked into a separate PR.

PR #351, opened 2022-09-17
Use ParallelDownloader for smaller files too

ppolewicz:


ppolewicz:
This mixes well with max part size selector implemented by Alex recently. With both his and this PR you can force a single part and still use ParallelDownloader to achieve higher single stream performance than what you'd get if there was no buffer between getting data from the cloud and writing it on the drive. ParallelDownloader will have a separate thread for retrieval and write(), so both of those can run concurrently (both release GIL)

b2sdk/transfer/inbound/downloader/parallel.py
```diff
@@ -64,7 +64,7 @@ def is_suitable(self, download_version: DownloadVersion, allow_seeking: bool):
             return False
         return self._get_number_of_streams(
             download_version.content_length
-        ) >= 2 and download_version.content_length >= 2 * self.min_part_size
+        ) >= 1 and download_version.content_length >= 2 * self.min_part_size
```
mpnowacki-reef, line 67: I think the first part, namely 
```
self._get_number_of_streams(
             download_version.content_length
         ) >= 1 
```
Is obsolete now, because if `self.max_workers` and `self._thread_pool._max_workers` are `None` it checks that `download_version.content_length >= self.min_part_size`, so it's a duplicate of `download_version.content_length >= 2 * self.min_part_size`. If `self._thread_pool._max_workers` or `self.max_workers` are not `None` then this comparison will only yield `False` if they are `0`. Should they ever be 0? It seems like a hack to disable parallel downloading that way, even though sdk users can mutate that parameter (https://b2-sdk-python.readthedocs.io/en/master/api/internal/transfer/inbound/downloader/parallel.html) - no sdk code ever actually sets it. `self._thread_pool._max_wrokers` is more easily settable to 0, but if that is supported behaviour, then I think `is_suitable` should be explicitly checking whether either of these values is 0 instead of doing redundant `mins`.

mpnowacki-reef (commented):
I think the logic can be simplified

ppolewicz:
I think we might want to wait for one more feature to land as it's going to add a new downloader. Then we can figure out how to do a new router for the downloaders.

mjurbanski-reef:
Superseded by #482

PR #352, opened 2022-09-17
Expose max streams parameter

ppolewicz:


b2sdk/api.py
```diff
@@ -64,6 +64,7 @@ def __init__(
         max_download_workers: Optional[int] = None,
         save_to_buffer_size: Optional[int] = None,
         check_download_hash: bool = True,
+        max_download_streams_per_file: Optional[int] = None,
```
mpnowacki-reef, line 67: how about an assertion to make sure it's higher than 1 if not none?

test/unit/v_all/test_api.py
```diff
@@ -22,6 +22,84 @@
...
+    def test_api_initialization(self):
+        class DummyA:
+            def __init__(self, *args, **kwargs):
+                pass
+
+        class DummyB:
+            def __init__(self, *args, **kwargs):
+                pass
+
+        for kwargs, _raw_api_class in [
```
mpnowacki-reef, line 40: any reason not do this via `parametrize`? purest doesn't support `with self.subTest` so asserts in loops in tests decrease readability, you have to guess which iteration failed from expected values. `parametrize` fixes that

mpnowacki-reef (commented):
see comments

PR #359, opened 2022-10-10
Context managers for performance metrics

kkalinowski-reef:
- context manager with constant memory usage replaced list (it's a performance downgrade, but readability upgrade);
- total moved into a context manager as well, class for stats handles all the logic now;
- data fetching moved from a `for`-loop to a generator `while`, to be able to enclose it in context manager easily.


ppolewicz:
The memory consumption is changed from O(N) to O(1), effectively from ~50MB * thread count down to <1KB * thread count.
The performance overhead has increased by ~300%, from ~1s of CPU to ~3s of CPU on a 10TB file, however the code is readable again and it is more important than a tiny performance hit.

PR #361, opened 2022-11-22
Add wildcard capabilities to Bucket.ls

kkalinowski-reef:
`Bucket.ls` got additional positional parameter `with_wildcard`. Setting it to True enables `*`, `?` and `[]` special operations on `folder_to_list`. Requires `recursive`.


b2sdk/bucket.py
```diff
@@ -346,33 +346,41 @@ def ls(
...
         :param fetch_count: how many entries to return or ``None`` to use the default. Acceptable values: 1 - 10000
         :param with_wildcard: Accepts "*", "?", "[]" and "[!]" in folder_to_list, similarly to what shell does.
-                              If ``True``, automatically assumes recursive was enabled
+                              As of 1.18.1 it can only be enabled when recursive is also enabled.
+                              Also, in this mode, folder_to_list is considered to be a filename or a pattern.
         :rtype: generator[tuple[b2sdk.v2.FileVersion, str]]
         :returns: generator of (file_version, folder_name) tuples
 
         .. note::
             In case of `recursive=True`, folder_name is returned only for first file in the folder.
```
ppolewicz, line 355: but if `with_wildcard=True`, `folder_name` is not returned. I agree that should be the case, but please document it.

test/unit/bucket/test_bucket.py
```diff
@@ -661,7 +665,27 @@ def test_matching_single_question_mark_named_file(self):
         ]
         actual = [
             (info.file_name, info.size, info.action, folder)
-            for (info, folder) in self.bucket_ls('b/?.txt', with_wildcard=True)
+            for (info, folder) in self.bucket_ls('b/?.txt', recursive=True, with_wildcard=True)
+        ]
+        self.assertEqual(expected, actual)
+
+    @pytest.mark.apiver(from_ver=2)
```
ppolewicz, line 672: it doesn't work on v1?

b2sdk/bucket.py
kkalinowski-reef: Actually, with `recursive=True` folder names are never returned. I've both clarified this comment and added test cases to explain the behaviour.

TL;DR:
```
b/1/test.txt
b/2/test.txt
b/3/test-1.txt
b/3/test-2.txt
```

If you perform `ls b/`, with recursive you'll get all the files, no folders. If you do it without recursive, you'll get first file from each folder, for a total of three results.

test/unit/bucket/test_bucket.py
kkalinowski-reef: `Bucket.ls` was not forwarding `kwargs`. Fixed.

ppolewicz (approved):

PR #362, opened 2022-11-22
Cache the bucket if the key is limited to one

kkalinowski-reef:
- checking whether there is a single bucket allowed for authorised key
- when the name of the bucket is missing, inform user about the issue


ppolewicz (approved):

ppolewicz (commented):
pep8 fix

PR #365, opened 2022-12-13
Removed information about replication being in closed beta

kkalinowski-reef:


ppolewicz:
leftover cleanup, good

Issue #368, opened 2022-12-25
[BUG] ValueError thrown when uploading local file with sub-millisecond mod time


ChangeinX:
**Environment**
Tester: ChangeinX
Version: b2sdk==1.18.0

**Expected Result**

What was wanted:

Upload a local file to the Backblaze bucket while preserving last modified time

**Actual Result**

What happened:

```
Traceback (most recent call last):
  File "C:\Users\ChangeinX\PycharmProjects\PJDB\lab\lambda_handler.py", line 54, in <module>
    main()
  File "C:\Users\ChangeinX\PycharmProjects\PJDB\lab\lambda_handler.py", line 50, in main
    upload_file(args.file_path)
  File "C:\Users\ChangeinX\PycharmProjects\PJDB\lab\lambda_handler.py", line 36, in upload_file
    bucket.upload_local_file(
  File "C:\Users\ChangeinX\PycharmProjects\PJDB\venv\lib\site-packages\logfury\_logfury\trace_call.py", line 86, in wrapper
    return function(*wrapee_args, **wrapee_kwargs)
  File "C:\Users\ChangeinX\PycharmProjects\PJDB\venv\lib\site-packages\b2sdk\bucket.py", line 505, in upload_local_file
    return self.upload(
  File "C:\Users\ChangeinX\PycharmProjects\PJDB\venv\lib\site-packages\logfury\_logfury\trace_call.py", line 86, in wrapper
    return function(*wrapee_args, **wrapee_kwargs)
  File "C:\Users\ChangeinX\PycharmProjects\PJDB\venv\lib\site-packages\b2sdk\bucket.py", line 551, in upload
    return self.create_file(
  File "C:\Users\ChangeinX\PycharmProjects\PJDB\venv\lib\site-packages\logfury\_logfury\trace_call.py", line 86, in wrapper
    return function(*wrapee_args, **wrapee_kwargs)
  File "C:\Users\ChangeinX\PycharmProjects\PJDB\venv\lib\site-packages\b2sdk\bucket.py", line 605, in create_file
    return self._create_file(
  File "C:\Users\ChangeinX\PycharmProjects\PJDB\venv\lib\site-packages\b2sdk\bucket.py", line 699, in _create_file
    return emerger_method(
  File "C:\Users\ChangeinX\PycharmProjects\PJDB\venv\lib\site-packages\logfury\_logfury\trace_call.py", line 86, in wrapper
    return function(*wrapee_args, **wrapee_kwargs)
[...]

PR #369, opened 2022-12-25
Content removed by author

ChangeinX:
Content removed by author for privacy.

### DIFF
--- a/b2sdk/file_version.py
+++ b/b2sdk/file_version.py
@@ -10,7 +10,6 @@
 
 from typing import Dict, Optional, Union, Tuple, TYPE_CHECKING
 import re
-import warnings
 from copy import deepcopy
 
 from .encryption.setting import EncryptionSetting, EncryptionSettingFactory
@@ -59,19 +58,19 @@ class BaseFileVersion:
     }
 
     def __init__(
-            self,
-            api: 'B2Api',
-            id_: str,
-            file_name: str,
-            size: int,
-            content_type: Optional[str],
-            content_sha1: Optional[str],
-            file_info: Dict[str, str],
-            upload_timestamp: int,
-            server_side_encryption: EncryptionSetting,
-            file_retention: FileRetentionSetting = NO_RETENTION_FILE_SETTING,
-            legal_hold: LegalHold = LegalHold.UNSET,
-            replication_status: Optional[ReplicationStatus] = None,
+        self,
+        api: 'B2Api',
+        id_: str,
+        file_name: str,
+        size: int,
+        content_type: Optional[str],
+        content_sha1: Optional[str],
+        file_info: Dict[str, str],
+        upload_timestamp: int,
+        server_side_encryption: EncryptionSetting,
+        file_retention: FileRetentionSetting = NO_RETENTION_FILE_SETTING,
+        legal_hold: LegalHold = LegalHold.UNSET,
+        replication_status: Optional[ReplicationStatus] = None,
     ):
         self.api = api
         self.id_ = id_
@@ -88,14 +87,12 @@ class BaseFileVersion:
 
         if SRC_LAST_MODIFIED_MILLIS in self.file_info:
             try:
-                self.mod_time_millis = int(self.file_info[SRC_LAST_MODIFIED_MILLIS])
-            except ValueError:
-                warnings.warn(
-                    'Invalid value for %s: %s' % (
-                        SRC_LAST_MODIFIED_MILLIS, self.file_info[SRC_LAST_MODIFIED_MILLIS]
-                    )
+                if SRC_LAST_MODIFIED_MILLIS in self.file_info:
+                    self.mod_time_millis = int(self.file_info[SRC_LAST_MODIFIED_MILLIS])
+            except ValueError as e:
+                raise ValueError(
+                    'Could not convert %s to int. %s' % (self.file_info[SRC_LAST_MODIFIED_MILLIS], e)
                 )
-                self.mod_time_millis = None
         else:
             self.mod_time_millis = self.upload_timestamp
 
@@ -188,9 +185,9 @@ class BaseFileVersion:
         return self._clone(legal_hold=legal_hold)
 
     def update_retention(
-            self,
-            file_retention: FileRetentionSetting,
-            bypass_governance: bool = False,
+        self,
+        file_retention: FileRetentionSetting,
+        bypass_governance: bool = False,
     ) -> 'BaseFileVersion':
         file_retention = self.api.update_file_retention(
             self.id_, self.file_name, file_retention, bypass_governance
@@ -231,23 +228,23 @@ class FileVersion(BaseFileVersion):
     ADVANCED_HEADERS_LIMIT = 2048
 
     def __init__(
-            self,
-            api: 'B2Api',
-            id_: str,
-            file_name: str,
-            size: Union[int, None, str],
-            content_type: Optional[str],
-            content_sha1: Optional[str],
-            file_info: Dict[str, str],
-            upload_timestamp: int,
-            account_id: str,
-            bucket_id: str,
-            action: str,
-            content_md5: Optional[str],
-            server_side_encryption: EncryptionSetting,
-            file_retention: FileRetentionSetting = NO_RETENTION_FILE_SETTING,
-            legal_hold: LegalHold = LegalHold.UNSET,
-            replication_status: Optional[ReplicationStatus] = None,
+        self,
+        api: 'B2Api',
+        id_: str,
+        file_name: str,
+        size: Union[int, None, str],
+        content_type: Optional[str],
+        content_sha1: Optional[str],
+        file_info: Dict[str, str],
+        upload_timestamp: int,
+        account_id: str,
+        bucket_id: str,
+        action: str,
+        content_md5: Optional[str],
+        server_side_encryption: EncryptionSetting,
+        file_retention: FileRetentionSetting = NO_RETENTION_FILE_SETTING,
+        legal_hold: LegalHold = LegalHold.UNSET,
+        replication_status: Optional[ReplicationStatus] = None,
     ):
         self.account_id = account_id
         self.bucket_id = bucket_id
@@ -301,10 +298,10 @@ class FileVersion(BaseFileVersion):
         return self.api.get_file_info(self.id_)
 
     def download(
-            self,
-            progress_listener: Optional[AbstractProgressListener] = None,
-            range_: Optional[Tuple[int, int]] = None,
-            encryption: Optional[EncryptionSetting] = None,
+        self,
+        progress_listener: Optional[AbstractProgressListener] = None,
+        range_: Optional[Tuple[int, int]] = None,
+        encryption: Optional[EncryptionSetting] = None,
     ) -> 'DownloadedFile':
         return self.api.download_file_by_id(
             self.id_,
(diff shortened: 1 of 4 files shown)


b2sdk/file_version.py
```diff
@@ -86,7 +86,13 @@ def __init__(
         self.replication_status = replication_status
 
         if SRC_LAST_MODIFIED_MILLIS in self.file_info:
-            self.mod_time_millis = int(self.file_info[SRC_LAST_MODIFIED_MILLIS])
+            try:
+                if SRC_LAST_MODIFIED_MILLIS in self.file_info:
+                    self.mod_time_millis = int(self.file_info[SRC_LAST_MODIFIED_MILLIS])
```
ppolewicz, line 91:
```suggestion
                self.mod_time_millis = int(self.file_info[SRC_LAST_MODIFIED_MILLIS])
```
the `if` statement was repeated from line 88

b2sdk/file_version.py
```diff
@@ -86,7 +86,13 @@ def __init__(
...
+            except ValueError as e:
+                raise ValueError(
+                    'Could not convert %s to int. %s' % (self.file_info[SRC_LAST_MODIFIED_MILLIS], e)
+                )
```
ppolewicz, line 95: This makes the program stop in some cases where we'd want it to keep running, such as if an incopatible client uploads a file with a non-int `src_last_modified_millis`. The affected bucket could then not be used by sdk/cli.

An alternative solution to this problem would be to, in such case:
- emit a warning
- try to trim the number to an integer, maybe that works
- failing to trim, just leave the field unset so that the sdk and cli can still use the bucket

b2sdk/utils/__init__.py
```diff
@@ -32,6 +32,12 @@ def b2_url_encode(s):
     :return: URL-encoded string
     :rtype: str
     """
+    # Ensure string is unicode
+    if not isinstance(s, str):
+        try:
+            s = str(s)
+        except TypeError as e:
+            raise TypeError('Cannot convert %r to unicode: %s' % (s, e))
```
ppolewicz, line 40: this change is not related to the other change so it should go to a different PR

ppolewicz (commented):
Hello, thank you for your contribution! There are a couple of minor issues with the pull request, some of which would be detected by the automated CI.

A test should be added for the added functionality to make sure your patch works after other changes done to this project in the future.

Also a changelog entry needs to be added. I suggest:
```
### Changed
* Handling non-int src_last_modified_millis better
```

b2sdk/file_version.py
ChangeinX: Thank you for the valuable feedback. A modification has been made to give a warning instead of throwing an exception. When uploading to a bucket, the behavior shows that the upload is successful then the exception is thrown. However, the alternative presented to emit a warning ensures that the desired result is maintained. 

Let me know if there's a different thought here, but a warning would be more useful since the current environment throws an exception, as shown in issue #368.

b2sdk/utils/__init__.py
ChangeinX: Thanks for the heads; I'll push this to a different PR.

ChangeinX:
Thank you for the feedback and swift response. Can you point me to or inform me about what is being sought after in a unit test? I see different approaches in the test suite and would like to contribute in a way that matches the expected standard. If there is no standard, may the unit tests contribute to the methodologies as proposed by Vladimir Khorikov?

b2sdk/file_version.py
```diff
@@ -58,19 +59,19 @@ class BaseFileVersion:
...
+            file_name: str,
+            size: int,
+            content_type: Optional[str],
+            content_sha1: Optional[str],
+            file_info: Dict[str, str],
+            upload_timestamp: int,
+            server_side_encryption: EncryptionSetting,
+            file_retention: FileRetentionSetting = NO_RETENTION_FILE_SETTING,
+            legal_hold: LegalHold = LegalHold.UNSET,
+            replication_status: Optional[ReplicationStatus] = None,
```
ppolewicz, line 74: our linter will not like that. Please don't reformat the code that is not related to the PR

b2sdk/utils/__init__.py
ppolewicz: Ok, let me know when you remove it from this PR

ppolewicz (changes requested):
I'm not familiar with this classification, will take a look at a lecture.

We use different type of tests for different purposes. Here I think it would be appropriate to make a test that simulates a situationh where a file with improper `mod_time_millis` format is being uploaded, listed and copied. Make sure the warning is emitted and that the value is set to None.

I think the most sophisticated part of the test framework is here: https://github.com/Backblaze/b2-sdk-python/tree/master/test/unit/v_all - it uses pytest, parametrize and conditions for different apiver versions where necessary

b2sdk/file_version.py
ChangeinX: My apologies. I did not notice this change. It may have occurred when running `nox -s format`

ChangeinX:
The book is a solid resource if you have an interest:

`Unit Testing Principles, Practices, and Patterns` by Vladimir Khorikov

The approach:

 - Unit tests should follow the AAA pattern
	- A sign of a bad test is one that multiple, arranges, acts, or asserts sections
		- A signal that the test is verifying more than one unit of behavior at once
			- If it does, split the test into multiple tests, one for each action

 - More than one line in the assert section signals a problem with the API for the system
   under test (SUT).
	- Making the client perform these actions together can lead to inconsistencies,
	  or *invariant violations* -- or code that hasn't been encapsulated

 - Identify the sut by naming it sut
 
 - Reuse test fixture initialization code by using factory methods and not passing
   it to the constructor
	- This type of reuse helps maintain a high degree of decoupling between tests
	  and improve readability.
	  
 - Don't use a strict naming policy on unit tests describing the behavior 
   and the name should reflect the behavior being tested
    - Name each test as if it were to be presented to a non-programmer
	  - separate words with underscores
	  - don't include the name of the method under test in the test name
	  
 - Parameterized tests reduce the code needed for similar tests
	- The drawback: tests become less readable as they are made more generic
	
 - Assertion libraries help further improve test readability by restructuring the 
   word order in assertions to read like plain English
   

Example:
```
class Calculator:
    def add(self, x, y):
        return x + y


class TestCalculator(Calculator):
    def test_add_two_numbers(self):
        # Arrange
        x = 1.5
        y = 2.5
        sut = Calculator()

        # Act
        result = sut.add(x, y)

        # Assert
        assert result == 4
```

b2sdk/utils/__init__.py
ChangeinX: It has been removed from the PR

b2sdk/file_version.py
ChangeinX: The invalid linting change has been reverted

ChangeinX (commented):
Content removed by author for privacy.

b2sdk/file_version.py
ppolewicz: This modification has been reverted and the program is now crashing again with a ValueError instead of emiting a warning.

ppolewicz (commented):
Thank you for performing the adjustments.

The changelog has still not been modified.

There is no test. I have seen half of a lecture about the book that you've mentioned and I generally agree with the presented principles, however in this particular case I think the most appropriate way is to add a single test which uses the api simulator to create a bucket, upload a file that has non-int mod_time_millis, then it tries to copy it, list the bucket and run some assertions. That single test will try to observe the behavior of the code in the conditions of our interest without looking into the low level implementation details. The test will be good until we add something really weird like alternative file-info storage (think a prefix of a end2end encrypted file), but in such case a full test review would be needed anyway.

Watch out for the file listing in old apiver - it used to return simple structures while now it returns FileVersion objects.

ChangeinX:
Further work is needed to resolve the conflict. 

Testing shows that specifying src_last_modified_millis exposes some issues with types and attributes
in potentially six different modules. 

Trimming appears to be the necessary approach for the entry point to recast as a string for encoding in
the method `b2_url_encode(s)` located at  `b2sdk/utils/__init__.py`

```
                if SRC_LAST_MODIFIED_MILLIS in self.file_info:
            try:
                self.mod_time_millis = int(self.file_info[SRC_LAST_MODIFIED_MILLIS].split('.')[0])
                self.mod_time_millis = str(self.mod_time_millis)
            except ValueError:
                warnings.warn(
                    'Invalid value for %s: %s' % (
                        SRC_LAST_MODIFIED_MILLIS, self.file_info[SRC_LAST_MODIFIED_MILLIS]
                    )
                )
                self.mod_time_millis = None
        else:
            self.mod_time_millis = self.upload_timestamp
```

Tested with: 
```
    def test_upload_with_time_attribute(self):
        bucket = self.bucket
        src_last_modified_millis = 1671500543.5804098

        sut = bucket.upload_bytes(
            b'test_file',
            'test_file',
            file_infos={'src_last_modified_millis': 1671500543.5804098}
        )

        assert isinstance(sut, VFileVersion)
```


Returns:

```
======================================================================================================= short test summary info ======================================================================================================= 
FAILED test/unit/bucket/test_bucket.py::TestDownloadDefault::test_download_update_mtime_v2 - TypeError: unsupported operand type(s) for /: 'str' and 'float'
FAILED test/unit/bucket/test_bucket.py::TestDownloadSimple::test_download_update_mtime_v2 - TypeError: unsupported operand type(s) for /: 'str' and 'float'
FAILED test/unit/bucket/test_bucket.py::TestDownloadParallel::test_download_update_mtime_v2 - TypeError: unsupported operand type(s) for /: 'str' and 'float'
FAILED test/unit/bucket/test_bucket.py::TestTruncatedDownloadParallel::test_download_update_mtime_v2 - TypeError: unsupported operand type(s) for /: 'str' and 'float'
FAILED test/unit/bucket/test_bucket.py::TestTruncatedDownloadSimple::test_download_update_mtime_v2 - TypeError: unsupported operand type(s) for /: 'str' and 'float'
FAILED test/unit/file_version/test_file_version.py::TestFileVersion::test_upload_with_time_attribute - AttributeError: 'float' object has no attribute 'encode'
======================================================================================= 6 failed, 490 passed, 50 skipped, 13 warnings in 9.28s ========================================================================================
```

ChangeinX:
Testing shows that the issue stems from the function b2_url_encode in `utils/__init__.py`
and not `b2sdk/file_version.py`

ChangeinX:
Thank you for working through this with me. This interaction will help me contribute better to the repo in the future.

ChangeinX:
Looking around the test suite, decoupling of tests, code encapsulation, and removal of branching may be in order. 

The test submitted passed for `v3` and `v2`, but as you mentioned, the return as FileVersion objects caused the test on v1 to fail. A fix for this would be to parameterize the tests with modules. The drawback to doing this, however, is that the tests potentially become less readable as they are made more generic.

test/unit/file_version/test_file_version.py
```diff
@@ -59,6 +59,19 @@ def test_format_ls_entry(self):
         )
         assert expected_entry == file_version_info.format_ls_entry()
 
+    @pytest.mark.apiver(from_ver=2, to_ver=3)
+    def test_upload_with_time_attribute_v2_v3(self):
```
ppolewicz, line 63: this test should not be restricted so much. While it might need to start from v2, it should probably not end on v3 unless we plan to write a different test / change the behavior in v4.

ppolewicz (changes requested):
Sorry for the delay, I've been buried under a pile of large volume review commits in the sdk for a while now.

It looks like the try/except blocks have been removed from the PR by accident.

ppolewicz:
This PR seems abandoned

PR #383, opened 2023-04-06
Support for cache control headers while uploading

przadka:
Added support for cache control parameter for uploading files
Modified tests and added two unit and one integration tests
There are a few places marked with FIXME because I was not sure whether changes are needed there.


b2sdk/bucket.py
```diff
@@ -493,6 +493,7 @@ def upload_bytes(
         encryption: Optional[EncryptionSetting] = None,
         file_retention: Optional[FileRetentionSetting] = None,
         legal_hold: Optional[LegalHold] = None,
+        cache_control: Optional[str] = None,
```
ppolewicz, line 496: if someone called it like this:

```
bucket.upload_bytes(
        self,
        data_bytes,
        file_name,
        content_type,
        file_infos,
        progress_listener,
        encryption,
        file_retention,
        legal_hold,
        large_file_sha1,
)
```
And you add `cache_control` as non-last argument, then the caller will suddenly find his positional argument to mis-align and fail due to a sha1 being accidentally put into the new `cache_control` variable instead of the old `large_file_sha1`. After your change every user of the sdk would have to adjust their code to make sure this doesn't happen. We would need to release a major version, 2.0.0.

Instead, please put the new argument at the very end of the argument list.

(if someone is using named arguments, that would still work, but we can't guarantee that every usr of the library does that)

b2sdk/bucket.py
```diff
@@ -792,6 +801,7 @@ def create_file(
         :param b2sdk.v2.EncryptionSetting encryption: encryption settings (``None`` if unknown)
         :param b2sdk.v2.FileRetentionSetting file_retention: file retention setting
         :param bool legal_hold: legal hold setting
+        :param str,None cache_control: cache control for the new file, optional
```
ppolewicz, line 804: this doesn't really say much. What would one put there, for example? Maybe mention RFC 2616 too.

b2sdk/file_version.py
```diff
@@ -576,7 +592,7 @@ def from_response_headers(self, headers):
             content_length=content_length,
             content_language=headers.get('Content-Language'),
             expires=headers.get('Expires'),
-            cache_control=headers.get('Cache-Control'),
+            cache_control= cache_control,
```
ppolewicz, line 595:
```suggestion
            cache_control=cache_control,
```
Extra space violates pep8. This shows that you haven't ran `nox -s format` before submitting the PR and you should really run the entire `nox -s test-3.10` (or whichever python version you are using) to make sure code is formatted properly, unit tests pass and integration tests pass.

Then CI will run it again on many versions of python.

b2sdk/replication/monitoring.py
```diff
@@ -45,6 +45,7 @@ class ReplicationScanResult(AbstractScanResult):
     source_has_large_metadata: Optional[bool] = None
     source_has_file_retention: Optional[bool] = None
     source_has_legal_hold: Optional[bool] = None
+    #FIXME add cache_control?
```
ppolewicz, line 48: no need to add it here

b2sdk/replication/monitoring.py
```diff
@@ -77,6 +78,7 @@ def from_files(
                         source_file_version.file_retention is not NO_RETENTION_FILE_SETTING,
                     'source_has_legal_hold':
                         source_file_version.legal_hold is LegalHold.ON,
+                    # FIXME add cache_control?
```
ppolewicz, line 81: nor here

b2sdk/session.py
```diff
@@ -557,6 +563,14 @@ def update_file_retention(
...
         )
 
+    def update_file_cache_control(
+        self,
+        file_id,
+        file_name,
+        cache_control: str,
+    ):
+        pass
+
```
ppolewicz, line 573:
```suggestion
```
I don't think this is needed, is it?

b2sdk/transfer/emerge/executor.py
```diff
@@ -329,6 +342,11 @@ def _find_unfinished_file_by_plan_id(
                 # pass UNKNOWN file_retention here - but raw_api/server won't allow it
                 # and we don't check it here
                 continue
+
+            if cache_control is None or file_.cache_control != encryption:
```
ppolewicz, line 346:
```suggestion
            if cache_control is None or file_.cache_control != cache_control:
```
A test should have found that.
This slipping through to a pull request shows that we are missing an important test.

b2sdk/transfer/emerge/executor.py
```diff
@@ -409,6 +427,11 @@ def _match_unfinished_file_if_possible(
             if encryption is not None and encryption != file_.encryption:
                 logger.debug('Rejecting %s: encryption mismatch', file_.file_id)
                 continue
+            
+            # FIXME as above - shoule this be handled in the same way as encryption?
```
ppolewicz, line 431: yes, it's clearly a different upload intent than the original

b2sdk/v1/file_version.py
```diff
@@ -163,6 +166,7 @@ def from_response_headers(self, headers):
             server_side_encryption=v2.EncryptionSettingFactory.from_response_headers(headers),
             file_retention=v2.FileRetentionSetting.from_response_headers(headers),
             legal_hold=v2.LegalHold.from_response_headers(headers),
+            cache_control=headers['cache-control'],
```
ppolewicz, line 169:
```suggestion
            cache_control=headers['Cache-Control'],
```

test/integration/test_raw_api.py
```diff
@@ -362,6 +362,23 @@ def raw_api_test_helper(raw_api, should_cleanup_old_buckets):
...
+        upload_auth_token,
+        file_name,
+        len(file_contents),
+        'text/plain',
+        file_sha1,
+        {'color': 'blue'},
+        io.BytesIO(file_contents),
+        cache_control='private, max-age=2222',
+    )
+
```
ppolewicz, line 381: in the list_file_versions test below we should check whether the value is saved properly and returned to us by the server in a way that we expect

test/unit/bucket/test_bucket.py
```diff
@@ -1402,6 +1402,20 @@ def test_upload_local_file_retention(self):
...
+        self._check_file_contents('file1', data)
+        self.assertEqual(file_info.cache_control, 'max-age=3600')
+
+    def test_upload_file_cache_control(self):
+        data = b'hello world'
+        file_info = self.bucket.upload_bytes(data, 'file1', cache_control='max-age=3600')
+        self.assertTrue(isinstance(file_info, VFileVersionInfo))
+        self._check_file_contents('file1', data)
+        self.assertEqual(file_info.cache_control, 'max-age=3600')
+
```
ppolewicz, line 1418: these tests are nearly identical - please refactor them in such a way that it doesn't cause so much code duplication

test/unit/bucket/test_bucket.py
```diff
@@ -1402,6 +1402,20 @@ def test_upload_local_file_retention(self):
...
+    def test_upload_bytes_cache_control(self):
+        data = b'hello world'
+        file_info = self.bucket.upload_bytes(data, 'file1', cache_control='max-age=3600')
+        self.assertTrue(isinstance(file_info, VFileVersionInfo))
+        self._check_file_contents('file1', data)
+        self.assertEqual(file_info.cache_control, 'max-age=3600')
+
+    def test_upload_file_cache_control(self):
+        data = b'hello world'
+        file_info = self.bucket.upload_bytes(data, 'file1', cache_control='max-age=3600')
```
ppolewicz, line 1414: judging by the name of the test it should be `upload_local_file` and not `upload_bytes`. In v0 it was done well and here it wasn't, which indicates that you have not self-reviewed your PR. It's strongly recommended :)

ppolewicz (commented):
Many missing docstrings

b2sdk/transfer/emerge/executor.py
przadka: @ppolewicz can you please elaborate? I am not sure what you meant by this comment

ppolewicz: if the cache control setting is different than in the given unfinished file upload, then we should reject it as a candidate for continuation (emit a debug log and `continue`

przadka:
Feedback implemented

b2sdk/v1/file_version.py
```diff
@@ -42,6 +42,7 @@ def __init__(
         server_side_encryption: Optional[v2.EncryptionSetting] = None,
         file_retention: Optional[v2.FileRetentionSetting] = None,
         legal_hold: Optional[v2.LegalHold] = None,
+        cache_control: Optional[str] = None,
         api: Optional['v1api.B2Api'] = None,
```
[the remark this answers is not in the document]
przadka: Feedback implemented in the latest commit

ppolewicz (approved):

PR #386, opened 2023-04-09
drop arrow dependency

ewdurbin:
replace with a non-english locale tolerant standard library implementation.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -98,7 +98,7 @@ jobs:
       - name: Install a locale for test
         run: |
           sudo apt -y update
-          sudo apt install language-pack-de
+          sudo apt -y install language-pack-de
       - uses: actions/checkout@v3
         with:
           fetch-depth: 0
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -95,10 +95,11 @@ jobs:
           - os: "windows-latest"
             python-version: "pypy-3.8"
     steps:
-      - name: Install a locale for test
-        run: |
-          sudo apt -y update
-          sudo apt -y install language-pack-de
+      - name: Install unit-test apt dependencies
+        uses: awalsh128/cache-apt-pkgs-action@v1.3.0
+        with:
+          packages: language-pack-de
+          version: 1.0  # This should be update each time the packages above are modified
       - uses: actions/checkout@v3
         with:
           fetch-depth: 0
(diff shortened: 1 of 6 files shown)


ppolewicz:
LGTM, I think the only missing thing now is an entry in `CHANGELOG.md` (`### Infrastructure` section, as it doesn't really impact the API but more CI/build/installation process.

ppolewicz:
@ewdurbin unit tests have failed on Ubuntu due to unsupported locale

.github/workflows/ci.yml
```diff
@@ -95,6 +95,10 @@ jobs:
           - os: "windows-latest"
             python-version: "pypy-3.8"
     steps:
+      - name: Install a locale for test
+        run: |
+          sudo apt -y update
+          sudo apt install language-pack-de
```
ppolewicz: that will take a long time and will cause the cache of CI to swell. Is there a different locale that is supported on all platforms without extra installation?

ewdurbin: No, by default ubuntu ships pretty minimal:

```
chungus:~ ee$ docker run --rm -it ubuntu bash
root@8d6202f9599e:/# locale -a
C
C.utf8
POSIX
```

Not sure what might be available in GitHub actions.

ewdurbin: 1) I'm not certain it will take a "long time"
2) It doesn't look like apt is being cached

Would you mind letting it run to determine if this works, and see what the time penalty is?

ewdurbin: It's certainly not free,

locally, `time docker run --rm -it ubuntu bash -c "apt update -y && apt install -y language-pack-de"` takes ~13s

ewdurbin: https://github.com/Backblaze/b2-sdk-python/pull/386/commits/9825eb9dbe7be3afee07258c14deb3381a82c18f should settle both concerns.

ppolewicz:
The unit test just needs to check if the context manager does what it should and, as you say, ubuntu has three locales it ships with. Can we just use these and not install any extra stuff?

ewdurbin:
> The unit test just needs to check if the context manager does what it should and, as you say, ubuntu has three locales it ships with. Can we just use these and not install any extra stuff?

That's certainly an option. I'll remove the test from ClockSkew and test the context manager in isolation.

ewdurbin:
@ppolewicz I'm done working on this pull request now. If you'd like to see this merged, please feel welcome to take over.

ppolewicz (approved):

ppolewicz:
Something silly is happening with the locales in tests, I'll take a look at it soon and get it in. Arrow has caused so much trouble in the past that I'd be happy to part ways with the dependency.

PR #388, opened 2023-04-11
Add support for custom upload timestamp

ppolewicz:


test/unit/bucket/test_bucket.py
```diff
@@ -1480,6 +1480,16 @@ def test_upload_local_file_incremental(self):
 
                 last_data = data
 
+    def assertBucketContents(self, expected, *args, **kwargs):
```
kkalinowski-reef, line 1483: This function is unused despite being just added. Did you mean to use it instead of `ls` function in `test_custom_timestamp`?

kkalinowski-reef (commented):
Would be nice to have also a test for unfinished file being matched or not, depending on custom upload timestamp.

test/unit/bucket/test_bucket.py
ppolewicz: good point

test/unit/bucket/test_bucket.py
```diff
@@ -1480,6 +1480,16 @@ def test_upload_local_file_incremental(self):
...
+        """
+        *args and **kwargs are passed to self.bucket_ls()
+        """
+        actual = [
+            (info.file_name, info.size, info.action, folder)
+            for (info, folder) in self.bucket_ls(*args, **kwargs)
+        ]
+        self.assertEqual(expected, actual)
+
```
ppolewicz, line 1492:
```suggestion
```

ppolewicz:
Reviewed by @kkalinowski-reef

PR #389, opened 2023-04-16
Remove arrow dependency

ppolewicz:
Closes #386


ppolewicz:
approved by Krzysztoferson

PR #390, opened 2023-04-27
Fixing circular symlinks (issue 513)

przadka:
Capped symlinks visits while scanning
New tests added
Added pytest-timeout to test requirements

### DIFF
--- a/test/unit/scan/test_folder_traversal.py
+++ b/test/unit/scan/test_folder_traversal.py
@@ -78,7 +78,7 @@ class TestFolderTraversal:
 
     @pytest.mark.skipif(
         platform.system() == 'Windows' and platform.python_implementation() == 'PyPy',
-        reason="Symlinks not supported on PyPy/Windows"
+        reason="Symlinks not supported on PyPy/Windows",
     )
     def test_folder_with_symlink_to_file(self, tmp_path):
 
@@ -108,7 +108,7 @@ class TestFolderTraversal:
 
     @pytest.mark.skipif(
         platform.system() == 'Windows' and platform.python_implementation() == 'PyPy',
-        reason="Symlinks not supported on PyPy/Windows"
+        reason="Symlinks not supported on PyPy/Windows",
     )
     @pytest.mark.timeout(5)
     def test_folder_with_circular_symlink(self, tmp_path):
@@ -136,7 +136,7 @@ class TestFolderTraversal:
 
     @pytest.mark.skipif(
         platform.system() == 'Windows' and platform.python_implementation() == 'PyPy',
-        reason="Symlinks not supported on PyPy/Windows"
+        reason="Symlinks not supported on PyPy/Windows",
     )
     @pytest.mark.timeout(5)
     def test_folder_with_symlink_to_parent(self, tmp_path):
@@ -177,7 +177,7 @@ class TestFolderTraversal:
 
     @pytest.mark.skipif(
         platform.system() == 'Windows' and platform.python_implementation() == 'PyPy',
-        reason="Symlinks not supported on PyPy/Windows"
+        reason="Symlinks not supported on PyPy/Windows",
     )
     @pytest.mark.timeout(5)
     def test_root_short_loop(self, tmp_path):
@@ -198,16 +198,16 @@ class TestFolderTraversal:
 
     @pytest.mark.skipif(
         platform.system() == 'Windows' and platform.python_implementation() == 'PyPy',
-        reason="Symlinks not supported on PyPy/Windows"
+        reason="Symlinks not supported on PyPy/Windows",
     )
     @pytest.mark.timeout(5)
     def test_root_parent_loop(self, tmp_path):
 
         # Create a symlink that points to the parent of the initial scanning point
         # tmp_path
-        # ├── start
-        # │   ├── file.txt
-        # └   └── symlink -> tmp_path
+        # └── start
+        #     ├── file.txt
+        #     └── symlink -> tmp_path
 
         (tmp_path / "start").mkdir()
         (tmp_path / "start" / "file.txt").write_text("content")
@@ -225,7 +225,7 @@ class TestFolderTraversal:
 
     @pytest.mark.skipif(
         platform.system() == 'Windows' and platform.python_implementation() == 'PyPy',
-        reason="Symlinks not supported on PyPy/Windows"
+        reason="Symlinks not supported on PyPy/Windows",
     )
     def test_symlink_that_points_deeper(self, tmp_path):
 
@@ -266,7 +266,7 @@ class TestFolderTraversal:
             fix_windows_path_limit(str(tmp_path / "b" / "d" / "d.txt")),
             fix_windows_path_limit(str(tmp_path / "b" / "d" / "e" / "e.txt")),
             fix_windows_path_limit(str(tmp_path / "f" / "f.txt")),
-            fix_windows_path_limit(str(tmp_path / "symlink/e.txt")),
+            fix_windows_path_limit(str(tmp_path / "symlink" / "e.txt")),
         ]
 
     @pytest.mark.skipif(
@@ -307,7 +307,7 @@ class TestFolderTraversal:
             fix_windows_path_limit(str(tmp_path / "b" / "c" / "c.txt")),
             fix_windows_path_limit(str(tmp_path / "b" / "d" / "d.txt")),
             fix_windows_path_limit(str(tmp_path / "b" / "d" / "e" / "e.txt")),
-            fix_windows_path_limit(str(tmp_path / "b" / "d" / "e" / "symlink/a.txt")),
+            fix_windows_path_limit(str(tmp_path / "b" / "d" / "e" / "symlink" / "a.txt")),
         ]
 
     @pytest.mark.skipif(
@@ -349,11 +349,10 @@ class TestFolderTraversal:
 
     @pytest.mark.skipif(
         platform.system() == 'Windows' and platform.python_implementation() == 'PyPy',
-        reason="Symlinks not supported on PyPy/Windows"
+        reason="Symlinks not supported on PyPy/Windows",
     )
     def test_valid_symlink_pattern_where_the_link_goes_down_and_up(self, tmp_path):
 
-        # Create a directory structure with a valid symlink pattern where the link goes down and up
         # tmp_path
         # ├── a
         # │   └── a.txt
@@ -395,7 +394,7 @@ class TestFolderTraversal:
 
     @pytest.mark.skipif(
         platform.system() == 'Windows' and platform.python_implementation() == 'PyPy',
-        reason="Symlinks not supported on PyPy/Windows"
+        reason="Symlinks not supported on PyPy/Windows",
     )
     def test_valid_symlink_pattern_where_the_link_goes_up_and_down(self, tmp_path):
 
@@ -435,7 +434,7 @@ class TestFolderTraversal:
 
     @pytest.mark.skipif(
         platform.system() == 'Windows' and platform.python_implementation() == 'PyPy',
-        reason="Symlinks not supported on PyPy/Windows"
+        reason="Symlinks not supported on PyPy/Windows",
     )
     @pytest.mark.timeout(5)
     def test_loop_that_goes_down_and_up(self, tmp_path):


przadka:
@ppolewicz After some additional time debugging symlinks on Windows this PR is ready for review. Some tests are skipped on Windows with PyPy, because PyPy do not support symlinks on Windows. I did a self-review before writing this comment. There are still some minor FIXMEs and I would appreciate some feedback on them.

b2sdk/scan/folder.py
```diff
@@ -204,6 +232,32 @@ def _walk_relative_paths(
...
+            real_path = os.readlink(absolute_path)
+
+            if is_ancestor_directory(local_dir, starting_dir):
+                # Skip symlink if it points to an ancestor directory of the initial scan directory
+                return
+
+            symlink_count = visited_symlinks.get(real_path, 0)
+            if symlink_count >= max_symlink_visits:
+                # Maximum symlink visits reached, stop the recursion
+                return
```
ppolewicz, line 258: This will silently damage backup jobs of everyone who is currently backing up a structure which contains data deeper than 10 symlinks in. Sure, it sounds like people shouldn't be doing anything like that but if they do, we don't want them to suddenly stop backing up a portion of their data when they update `b2-sdk-python` by 0.1.0.

BTW we should have a test that shows that data deeper than the threshold is still backed up properly though a bit slower because when we are so deep we are testing everything for circular symlinks which involves more syscalls than normally. That potential slowness should be, in turn, described in the docstring of the class.

b2sdk/scan/folder.py
```diff
@@ -204,6 +232,32 @@ def _walk_relative_paths(
...
+
+        if os.path.islink(local_dir):
+            absolute_path = os.path.abspath(local_dir)
+            real_path = os.readlink(absolute_path)
+
+            if is_ancestor_directory(local_dir, starting_dir):
+                # Skip symlink if it points to an ancestor directory of the initial scan directory
+                return
+
+            symlink_count = visited_symlinks.get(real_path, 0)
```
ppolewicz, line 255: this wouldn't be needed if `visited_symlinks` would be `defaultdict(int)` or `collections.Counter` (the former seems more appropriate here)

b2sdk/scan/policies.py
```diff
@@ -140,6 +141,7 @@ def __init__(
         :param exclude_modified_after: optionally exclude file versions (both local and b2) modified after (in millis)
         :param exclude_uploaded_before: optionally exclude b2 file versions uploaded before (in millis)
         :param exclude_uploaded_after: optionally exclude b2 file versions uploaded after (in millis)
+        :param symlink_max_visits: maximum number of times to follow a symlink, to avoid infinite loops (default: 10)
```
ppolewicz, line 144: I don't think users of this class want us to limit this. I think they just want a warning and automatic recovery (skip) whenever an infinite symlink loop is encountered.

noxfile.py
```diff
@@ -39,6 +39,7 @@
     'pytest-lazy-fixture==0.6.3',
     'pyfakefs==4.5.6',
     'pytest-xdist==2.5.0',
+    'pytest-timeout==2.1.0',
```
ppolewicz, line 42: I don't think we need this, do we?

test/unit/scan/test_scan_policies.py
```diff
@@ -7,14 +7,207 @@
...
 #
 ######################################################################
+from pathlib import Path
+import platform
 import re
-
+from unittest.mock import MagicMock
+from b2sdk.scan.folder import LocalFolder
+from b2sdk.utils import fix_windows_path_limit
 import pytest
```
ppolewicz, line 16: imports are not grouped properly as `isort` would group them. As a convention in this project we group imports manually (should probably just use `isort` - I'm not sure why we don't have it set up right now, maybe it was clashing with yapf?)

test/unit/scan/test_scan_policies.py
```diff
@@ -7,14 +7,207 @@
...
 
+class TestFolderTraversal:
+    def test_simple_folder(self, tmp_path):
+        d = tmp_path / "dir"
+        d.mkdir()
+        (d / "file1.txt").write_text("content1")
+        (d / "file2.txt").write_text("content2")
+        (d / "file3.txt").write_text("content3")
+
+        folder = LocalFolder(str(d))
```
ppolewicz, line 30: generally we should use Pathlib everywhere. `b2sdk.v2` might use string paths in some places, but it's a legacy problem and I think switching to pathlib in `v3` is a good direction (though maybe lets not do it in this PR).

test/unit/scan/test_scan_policies.py
```diff
@@ -7,14 +7,207 @@
...
+        local_paths = folder.all_files(reporter=MagicMock())
+        absolute_paths = [path.absolute_path for path in list(local_paths)]
+
+        assert len(absolute_paths) == 3
+        assert absolute_paths == [
+            fix_windows_path_limit(str(d / "file1.txt")),
```
ppolewicz, line 36: I think it's time for `fix_windows_path_limit` to learn how to work with Pathlib arguments

test/unit/scan/test_scan_policies.py
```diff
@@ -7,14 +7,207 @@
...
+from unittest.mock import MagicMock
+from b2sdk.scan.folder import LocalFolder
+from b2sdk.utils import fix_windows_path_limit
 import pytest
 
 from apiver_deps import ScanPoliciesManager
 from apiver_deps_exception import InvalidArgument
 
 
+class TestFolderTraversal:
```
ppolewicz, line 22: does this belong in `test_scan_policies.py`? I think we might need a new file like `test_folder_traversal.py`or something

test/unit/scan/test_scan_policies.py
```diff
@@ -7,14 +7,207 @@
...
+    def test_simple_folder(self, tmp_path):
```
ppolewicz, line 23:
```suggestion
    def test_flat_folder(self, tmp_path):
```

test/unit/scan/test_scan_policies.py
```diff
@@ -7,14 +7,207 @@
...
+            fix_windows_path_limit(str(d1 / "file1.txt")),
+            fix_windows_path_limit(str(d1 / "file2.txt")),
+            fix_windows_path_limit(str(d2 / "file3.txt")),
+            fix_windows_path_limit(str(d2 / "file4.txt"))
+        ]
+
+    @pytest.mark.skipif(
+        platform.system() == 'Windows' and platform.python_implementation() == 'PyPy',
+        reason="Symlinks not supported on PyPy/Windows"
+    )
```
ppolewicz, line 67: 👍 
good stuff

test/unit/scan/test_scan_policies.py
```diff
@@ -7,14 +7,207 @@
...
+        d2 = tmp_path / "dir2"
+        d2.mkdir()
+        (d2 / "file3.txt").write_text("content3")
+        (d2 / "file4.txt").write_text("content4")
+
+        folder = LocalFolder(str(tmp_path))
+        local_paths = folder.all_files(reporter=MagicMock())
+        absolute_paths = [path.absolute_path for path in list(local_paths)]
+
+        assert len(absolute_paths) == 4
```
ppolewicz, line 56:
```suggestion
```
the assert below already checks list length

test/unit/scan/test_scan_policies.py
```diff
@@ -7,14 +7,207 @@
...
+        d.mkdir()
+        (d / "file1.txt").write_text("content1")
+        (d / "file2.txt").write_text("content2")
+        (d / "file3.txt").write_text("content3")
+
+        folder = LocalFolder(str(d))
+        local_paths = folder.all_files(reporter=MagicMock())
+        absolute_paths = [path.absolute_path for path in list(local_paths)]
+
+        assert len(absolute_paths) == 3
```
ppolewicz, line 34:
```suggestion
```
the assert below already checks list length. Moreover if this assert fails, you'll get something like `3 != 2` which is way less useful than the feedback from the assert below.

test/unit/scan/test_scan_policies.py
```diff
@@ -7,14 +7,207 @@
...
+        assert absolute_paths == [
+            fix_windows_path_limit(str(d / "file1.txt")),
+            fix_windows_path_limit(str(d / "file2.txt")),
+            fix_windows_path_limit(str(d / "file3.txt"))
```
ppolewicz, line 38:
```suggestion
            fix_windows_path_limit(str(d / "file3.txt")),
```

test/unit/scan/test_scan_policies.py
```diff
@@ -7,14 +7,207 @@
...
+        folder = LocalFolder(str(tmp_path))
+        local_paths = folder.all_files(reporter=MagicMock())
+        absolute_paths = [path.absolute_path for path in list(local_paths)]
+
+        assert len(absolute_paths) == 4
+        assert absolute_paths == [
+            fix_windows_path_limit(str(d1 / "file1.txt")),
+            fix_windows_path_limit(str(d1 / "file2.txt")),
+            fix_windows_path_limit(str(d2 / "file3.txt")),
+            fix_windows_path_limit(str(d2 / "file4.txt"))
```
ppolewicz, line 61:
```suggestion
            fix_windows_path_limit(str(d2 / "file4.txt")),
```
self-review should have caught that - you have a similar thing in a few places and you use trailing commas inconsistently... or we should install something that will fix it automatically, ha ha.

test/unit/scan/test_scan_policies.py
```diff
@@ -7,14 +7,207 @@
...
+
+    #FIXME the following two tests could be combined to avoid code duplication
+    # but I decide to keep them separate for now to make it easier to debug
+
+    @pytest.mark.skipif(
+        platform.system() == 'Windows' and platform.python_implementation() == 'PyPy',
+        reason="Symlinks not supported on PyPy/Windows"
+    )
+    @pytest.mark.timeout(5)  # Set a 5-second timeout for this test
+    def test_folder_with_circular_symlink(self, tmp_path):
```
ppolewicz, line 97: I think we should recurse down 10 levels (or whatever the limit is) and then realize that we are in an infinite loop, raise an exception, catch it upstairs in the stack and see if the place where we are is an infinite loop already or have we went over it already and we are in a healthy directory again. If we are in a loop, raise again. Eventually we'll come all the way to the top where the _walk function fell into the loop and we'll realize that was a mistake - then we'll emit a warning (maybe tell the reporter what happened?) and skip.

If we wouldn't do that, we'd (as shown in this test) upload the same file (or files) ~10 times, wasting capacity. I do not think this is what the users of this class would want.

test/unit/scan/test_scan_policies.py
```diff
@@ -7,14 +7,207 @@
...
+
+        symlink_file = tmp_path / "symlink_file.txt"
+        symlink_file.symlink_to(file)
+
+        folder = LocalFolder(str(tmp_path))
+        local_paths = folder.all_files(reporter=MagicMock())
+
+        absolute_paths = [path.absolute_path for path in list(local_paths)]
+
+        assert len(absolute_paths) == 2
```
ppolewicz, line 83:
```suggestion
```
the assert below already checks list length

test/unit/scan/test_scan_policies.py
```diff
@@ -7,14 +7,207 @@
...
+
+        folder = LocalFolder(str(tmp_path))
+
+        policies_manager = ScanPoliciesManager()
+        policies_manager.max_symlink_visits = 3
+
+        local_paths = folder.all_files(reporter=MagicMock(), policies_manager=policies_manager)
+        absolute_paths = [path.absolute_path for path in list(local_paths)]
+
+        assert len(absolute_paths) == 4
```
ppolewicz, line 114:
```suggestion
```
the assert below already checks list length

test/unit/scan/test_scan_policies.py
```diff
@@ -7,14 +7,207 @@
...
+        symlink_dir.symlink_to(d)
+
+        policies_manager = ScanPoliciesManager()
+        policies_manager.max_symlink_visits = 1
+
+        folder = LocalFolder(str(tmp_path))
+        local_paths = folder.all_files(reporter=MagicMock(), policies_manager=policies_manager)
+        absolute_paths = [path.absolute_path for path in list(local_paths)]
+
+        assert len(absolute_paths) == 2
```
ppolewicz, line 145:
```suggestion
        assert len(absolute_paths) == 2
```
the assert below already checks list length. Moreover if this assert fails, you'll get something like `3 != 2` which is way less useful than the feedback from the assert below.

test/unit/scan/test_scan_policies.py
```diff
@@ -7,14 +7,207 @@
...
+
+        folder = LocalFolder(str(tmp_path))
+
+        policies_manager = ScanPoliciesManager()
+        policies_manager.max_symlink_visits = 3
+
+        local_paths = folder.all_files(reporter=MagicMock(), policies_manager=policies_manager)
+        absolute_paths = [path.absolute_path for path in list(local_paths)]
+
+        assert len(absolute_paths) == 6
```
ppolewicz, line 196:
```suggestion
```
the assert below already checks list length. Moreover if this assert fails, you'll get something like `3 != 2` which is way less useful than the feedback from the assert below.

test/unit/scan/test_scan_policies.py
```diff
@@ -7,14 +7,207 @@
...
+
+    # Just for fun, another test, based on a comment by @zackse available here:
+    # https://github.com/Backblaze/B2_Command_Line_Tool/issues/513#issuecomment-426751216
+
+    @pytest.mark.skipif(
+        platform.system() == 'Windows' and platform.python_implementation() == 'PyPy',
+        reason="Symlinks not supported on PyPy/Windows"
+    )
+    @pytest.mark.timeout(5)  # Set a 5-second timeout for this test
+    def test_circular_interesting_case(self, tmp_path):
```
ppolewicz, line 159: this test needs a docstring that explains the situation so that I don't have to read every single line of it to understand what is going on

b2sdk/scan/folder.py
```diff
@@ -178,15 +178,43 @@ def ensure_non_empty(self):
...
+        starting_dir: str = None,
     ):
         """
         Yield a File object for each of the files anywhere under this folder, in the
         order they would appear in B2, unless the path is excluded by policies manager.
 
         :param relative_dir_path: the path of this dir relative to the scan point, or '' if at scan point
         """
+
+        def is_ancestor_directory(potential_ancestor, dir):
```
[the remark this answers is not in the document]
przadka: @ppolewicz  _walk_relative_paths uses string patha and this is why I decided to use string path arguments for this helper function. Do you recommend to use Path objects here? in that case this function simply becomes:

`
return potential_ancestor in potential_descendant.parents
`

so probably not worth to do it as a separate function. At the same time, converting string paths to Path objects seems clumsy and inconsistent. I would prefer do refactor _walk_relative_paths to use Path objects instead of path strings. What do you think?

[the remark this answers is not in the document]
ppolewicz: Since `_walk_relative_paths` is private, you should refactor it to use path. Generally pathlib is in the standard library now and we should use it when appropriate (which usually means we have to move away from string paths - and that's good)

b2sdk/scan/folder.py
```diff
@@ -178,15 +178,43 @@ def ensure_non_empty(self):
...
+            # e.g. /home/user/ancestor is an ancestor of /home/user/ancestor/descendant
+
+            potential_ancestor = os.path.abspath(os.path.realpath(potential_ancestor))
+            dir = os.path.abspath(os.path.realpath(dir))
+
+            return os.path.commonpath([potential_ancestor]) == os.path.commonpath(
+                [potential_ancestor, dir]
+            )
+
+        #FIXME should I construct starting_dir dynamically based on relative_dir_path and local_dir?
```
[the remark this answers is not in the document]
przadka: @ppolewicz This parameter represents the original starting point of the directory scan. It is required to catch a situation described by @zackse in his comment here: https://github.com/Backblaze/B2_Command_Line_Tool/issues/513#issuecomment-426751216

While scanning the directory tree from the initial scanning point, we may jump outside of the original initial scanning point via symlinks. If the linked/target directory contains the initial scanning point, then we consider it a bad link.

Maybe this is not the correct implementation... After your comment, I am now thinking that maybe we should only rely on symlink visit counter to skip symlinks. In the current implementation I mixed the two approaches:

* I skip symlinks which are pointing to ancestor of the initial scan point
* I skip symlinks if we reach the symlink visist limit

I guess the second one is more general and covers the first one as a special case of sorts. But it will take more time to discover that we are actually in a circular symlinks case. Do you have any suggestions on what would be a better approach?

b2sdk/scan/folder.py
```diff
@@ -204,6 +232,32 @@ def _walk_relative_paths(
...
         # This is because in Unicode '.' comes before '/', which comes before '0'.
         names = []  # list of (name, local_path, relative_file_path)
+
+        if starting_dir is None:
+            starting_dir = local_dir
+
+        # Set max_symlink_visits based on the policies_manager object
+        # Fall back to 10 if max_symlink_visits is not set
+        # FIXME should we use a default value from somewhere else?
+        max_symlink_visits = getattr(policies_manager, 'max_symlink_visits', 10)
```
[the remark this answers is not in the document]
przadka: @ppolewicz did you forget to include a comment here? Because the next comment seems to be a continuation of something I don't see.

b2sdk/scan/folder.py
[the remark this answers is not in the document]
przadka: refactored and removed starting_dir parameter completely

b2sdk/scan/folder.py
```diff
@@ -215,26 +213,19 @@ def _walk_relative_paths(
...
             real_path = local_dir.resolve()
-
-            if real_path in starting_dir.parents:
-                # Skip symlink if it points to an ancestor directory of the initial scan directory
-                #FIXME I think I should raise an error here instead of skipping the symlink
-                return
-
-            if real_path in visited_symlinks:
-                # Infinite symlink loop detected, skip symlink
+            inode_number = real_path.stat().st_ino
```
przadka, line 221: @ppolewicz As per our discussion, I changed the approach to reduce memory usage. Assuming:

- 64-bit system, with 28 bytes per integer
- 1 billion directories in the scanned directory tree
-  5% of these are symlinks (50m)

The pessimistic size of the visited_symlinks would be around 1.4GB. I would argue this is acceptable, but please share your thoughts.

we are on 64-bit system and an integer takes 28 bytes, we can calculate have the following: 1 billions folders in the traversed tree, with

test/unit/scan/test_folder_traversal.py
```diff
@@ -0,0 +1,195 @@
...
+            fix_windows_path_limit(
+                str(
+                    tmp_path / "parent" / "child" / "grandchild" / "symlink_dir" / "child" /
+                    "grandchild" / "file5.txt"
+                )
+            ),
+            fix_windows_path_limit(
+                str(tmp_path / "parent" / "child" / "grandchild" / "symlink_dir" / "file3.txt")
+            ),
+        ]
```
ppolewicz, line 195:
```suggestion
        ]  # yapf: disable
```
and format the lines to be long or it's made unreadable by the line wrapper

b2sdk/scan/report.py
```diff
@@ -192,6 +192,17 @@ def local_permission_error(self, path):
     def symlink_skipped(self, path):
         pass
 
+    def circular_symlink_skipped(self, path):
+        """
+        Add a circular symlink error message to the list of warnings.
+
+        :param path: file path
+        :type path: str
```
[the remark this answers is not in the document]
przadka: ok, i will fix this for all the methods in this class

b2sdk/scan/report.py
```diff
@@ -192,6 +192,17 @@ def local_permission_error(self, path):
     def symlink_skipped(self, path):
         pass
 
+    def circular_symlink_skipped(self, path):
```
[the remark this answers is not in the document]
przadka: ok, i will fix this for all the methods in this class

test/unit/scan/test_folder_traversal.py
```diff
@@ -0,0 +1,195 @@
...
+        assert absolute_paths == [
+            fix_windows_path_limit(str(file)),
+            fix_windows_path_limit(str(symlink_file))
+        ]
+
+    @pytest.mark.skipif(
+        platform.system() == 'Windows' and platform.python_implementation() == 'PyPy',
+        reason="Symlinks not supported on PyPy/Windows"
+    )
+    @pytest.mark.timeout(5)  # Set a 5-second timeout for this test
```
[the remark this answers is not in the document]
przadka: @kkalinowski-reef  Yes, this is to avoid infinite loops. Given how the code work now, it is probably not necessary, but I wrote the tests first, so I wanted to make sure they finish running. I did not experience any issues with other plugins and CI on Windows runs normally. You think I should remove this?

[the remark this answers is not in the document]
przadka: @ppolewicz maybe you have some feedback on this?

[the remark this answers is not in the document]
ppolewicz: a test with an infinite loop is easy to detect with -vv, but this is handy and since it's a dev/test dependency, I'm ok with it

b2sdk/scan/folder.py
```diff
@@ -204,39 +212,52 @@ def _walk_relative_paths(
...
+        if local_dir.is_symlink():
+            real_path = local_dir.resolve()
+            inode_number = real_path.stat().st_ino
+            if inode_number in visited_symlinks:
+                # Infinite symlink loop detected, report warning and skip symlink
+                if reporter is not None:
+                    reporter.circular_symlink_skipped(str(local_dir))
+                return
+
+            visited_symlinks.add(inode_number)
```
ppolewicz, line 227: the first and last line of that range is an operation that needs to find out whether the set contains an item, the second actually adds it. I think for big sets it would be significantly faster to do `len(visited_symlinks)`, `visited_symlinks.add(inode_number)` and `len()` again to see if we succeeded in adding an element (because if we didn't then it was already there). A  great opportunity for you to use the `timeit` module!

b2sdk/scan/folder.py
```diff
@@ -204,39 +212,52 @@ def _walk_relative_paths(
...
+
+        for name in [x.name for x in local_dir.iterdir()]:
```
ppolewicz, line 229: this will eat more memory than it should when running in a really wide directory. You could 
```suggestion
        for name in (x.name for x in local_dir.iterdir()):
```
to prevent allocation of a list containing everything. With 1B files and 1KB file name that's like 1TB of memory for that list.

ppolewicz, line 229: this will eat more memory than it should when running in a really wide directory. You could 
```suggestion
        for name in (x.name for x in local_dir.iterdir()):
```
to prevent allocation of a list containing everything.

test/unit/scan/test_folder_traversal.py
```diff
@@ -0,0 +1,193 @@
...
+                )
+            ),
+            fix_windows_path_limit(
+                str(
+                    tmp_path / "parent" / "child" / "grandchild" / "symlink_dir" / "child" / "grandchild" / "file5.txt"
+                )
+            ),
+            fix_windows_path_limit(
+                str(tmp_path / "parent" / "child" / "grandchild" / "symlink_dir" / "file3.txt")
+            ),
```
ppolewicz, line 192:
```suggestion
            fix_windows_path_limit(str(tmp_path / "parent" / "child" / "grandchild" / "symlink_dir" / "child" / "file4.txt")),
            fix_windows_path_limit(str(tmp_path / "parent" / "child" / "grandchild" / "symlink_dir" / "child" / "grandchild" / "file5.txt")),
            fix_windows_path_limit(str(tmp_path / "parent" / "child" / "grandchild" / "symlink_dir" / "file3.txt")),
```

test/unit/scan/test_folder_traversal.py
```diff
@@ -0,0 +1,193 @@
...
+            fix_windows_path_limit(str(tmp_path / "dir" / "symlink_dir" / "file1.txt")),
+        ]
+
+    @pytest.mark.skipif(
+        platform.system() == 'Windows' and platform.python_implementation() == 'PyPy',
+        reason="Symlinks not supported on PyPy/Windows"
+    )
+    @pytest.mark.timeout(5)  # Set a 5-second timeout for this test
+    def test_folder_with_symlink_to_parent(self, tmp_path):
+
```
ppolewicz, line 146: Some tests should be added:

1. test_root_short_loop where `tmp_path` is a symlink to itself
2. test_root_parent_loop where `tmp_path` is a symlink to its parent
3. test_symlink_that_points_deeper, `a, b/c -> b/d/e, f`
4. test_symlink_that_poitns_up, `a, b/c/d -> b/e, f`
5. test_elaborate_infinite_loop (symlink to a symlink to a symlink to a symlink which points to the first symlink), `a, b -> c, c -> d, d -> e, e -> b, f`
6. test_valid_symlink_pattern_where_the_link_goes_down_and_up `a, b -> c/d, d -> e, f`
7. test_valid_symlink_pattern_where_the_link_goes_up_and_down `a, b/c -> d, d -> e/f, g`
8. test_loop_that_goes_down_and_up, `a, b -> c/d, d -> e, e -> b, f`
9. test_loop_that_goes_up_and_down `a, b/c -> d, d -> e/f, f -> b/c, g`

I think your implementation is not vulnerable to any issues that this test might detect, but the tests are also here so that when someone tries to "optimize" your code in the future then any issues with the new implementation will be caught by the test harness.

b2sdk/scan/folder.py
przadka: Nice one!

```
Time elapsed for original code: 4.718000127468258e-06
Time elapsed for optimized code: 2.958000550279394e-06
```

ppolewicz: 🤔 is the optimized code faster for an empty set too?

przadka: No, for smaller sets the oiriginal code is better or similar:

Empty set:
```
Time elapsed for original code: 1.319000034527562e-06
Time elapsed for optimized code: 2.3310000187848345e-06
```

1000 elements:
```
Time elapsed for original code: 1.0374999988016498e-05
Time elapsed for optimized code: 1.1050000011891825e-05
```

1m:
```
Time elapsed for original code: 5.107999982101319e-06
Time elapsed for optimized code: 2.774000108729524e-06
```

10m:
```
Time elapsed for original code: 6.407000000763219e-06
Time elapsed for optimized code: 3.318000040053448e-06
```

test/unit/scan/test_scan_policies.py
```diff
@@ -8,7 +8,6 @@
 #
 ######################################################################
 import re
-
```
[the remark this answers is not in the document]
przadka: Thanks for your feedback. The 're' import was a leftover, not needed for this test. I've removed it now.

test/unit/scan/test_folder_traversal.py
przadka: @ppolewicz what is the purpose of this test:

> test_root_parent_loop where tmp_path is a symlink to its parent

And what would you expect to see when running it?

przadka: @ppolewicz The newest version contains all the tests above apart from the one I asked about earlier. It would be great if you could (at least) review the directory structures in the comments to make sure I understood your intentions correctly. In the mean time, I need to debug why the new tests fail on Windows 🤯🤯🤯

ppolewicz: it's a symlink like `/home/user/aaa` -> `/home/user` and then we try to sync `/home/user/aaa` somewhere

test/unit/scan/test_scan_policies.py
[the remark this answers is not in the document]
ppolewicz: unused imports should be detected by CI (except the apiver dirs) - CI must be misconfigured somehow? Can you confirm this is happening?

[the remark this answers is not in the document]
przadka: Yes, i confirm this. I added 3 unused imports to my latest commit and CI is all green.

PR #392, opened 2023-05-17
chore: build python wheels as well

miketheman:
Doesn't yet use the modern PEP517 approach, since that's a much larger refactor in tooling and choices than adding the ability create wheels.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Infrastructure
 * Remove dependency from `arrow`
+* Build Python wheels for distribution
 
 ## [1.20.0] - 2023-03-23
 
--- a/noxfile.py
+++ b/noxfile.py
@@ -169,11 +169,11 @@ def cover(session):
 @nox.session(python=PYTHON_DEFAULT_VERSION)
 def build(session):
     """Build the distribution."""
-    # TODO: consider using wheel as well
     session.run('pip', 'install', *REQUIREMENTS_BUILD)
     session.run('python', 'setup.py', 'check', '--metadata', '--strict')
     session.run('rm', '-rf', 'build', 'dist', 'b2sdk.egg-info', external=True)
     session.run('python', 'setup.py', 'sdist', *session.posargs)
+    session.run('python', 'setup.py', 'bdist_wheel', *session.posargs)
 
     # Set outputs for GitHub Actions
     if CI:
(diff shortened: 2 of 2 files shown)


ppolewicz:
It doesn't work, but I agree that we should have wheels. @miketheman do you want to finish it or should I take over now?

miketheman:
@ppolewicz i think the build env needed the wheel package, try again?

mlech-reef:
Hi.
I am afraid this change will break Continues Deployment, particularly this step:

```yaml
      - name: Upload the distribution to GitHub
        uses: actions/upload-release-asset@v1
        with:
          upload_url: ${{ steps.create-release.outputs.upload_url }}
          asset_path: ${{ steps.build.outputs.asset_path }}
          asset_name: ${{ steps.build.outputs.asset_name }}
          asset_content_type: application/gzip
```

CD parameters like `asset_path`, `asset_name` are set in `noxfile.py`:
```python
    # Set outputs for GitHub Actions
    if CI:
        asset_path = glob('dist/*')[0]
        print('::set-output name=asset_path::', asset_path, sep='')

        asset_name = os.path.basename(asset_path)
        print('::set-output name=asset_name::', asset_name, sep='')

        version = os.environ['GITHUB_REF'].replace('refs/tags/v', '')
        print('::set-output name=version::', version, sep='')
```

With bdist_wheel enabled, `asset_path = glob('dist/*')[0]` will point to the wheel file:
```
>>> from glob import glob
>>> glob('dist/*')
['dist/b2sdk-1.21.1.dev16+g7f0e3234-py3-none-any.whl', 'dist/b2sdk-1.21.1.dev16+g7f0e3234.tar.gz']
```

If we want to keep two assets (sdist and bdist_wheel) we need to refactor this part as well, as now we are supporting only one. Another place of the Continues Deployment pipeline that should be tested is:
```yaml
      - name: Upload the distribution to PyPI
        if: ${{ env.B2_PYPI_PASSWORD != '' }}
        uses: pypa/gh-action-pypi-publish@v1.3.1
        with:
          user: __token__
          password: ${{ env.B2_PYPI_PASSWORD }}
```

We need to make sure that `pypa/gh-action-pypi-publish@v1.3.1` publish both distributions, not only one.

So introducing wheel require more changes, unless we replace sdist with bdist_wheel which should work out-of-the box I think, just need to change `asset_content_type: application/gzip` to `asset_content_type: application/x-wheel+zip`

mlech-reef:
Github Action `actions/upload-release-asset` used by this project for uploading assets is deprecated, so if we want to upload more than one asset, would be good to switch to `softprops/action-gh-release` which by the way is already used in https://github.com/Backblaze/B2_Command_Line_Tool. Uploading multiple assets is easier there which would probably simplify the `noxfile.py`

ppolewicz:
ok @mlech-reef please fix all of that and test it. Then we'll merge it.

PR #393, opened 2023-05-23
Remove dependency from tqdm

ppolewicz:
Projects like B2CLI should require it directly. The sdk generally works fine without that dependency which pulls its own dependencies and it is a problem for some projects. We'll continue to support classes that use tqdm.

Any existing installations should keep their dependency in their venv and new versions of the CLI as well as any other console applications should depend on tqdm directly, if they choose to use it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 * Circular symlinks no longer cause infinite loops when syncing a folder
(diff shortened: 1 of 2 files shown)


ppolewicz:
Ok so the problem is that TqdmProgressReporter was (by mistake) a part of the public interface so now removing the dependency will potentially break the clients using that class (if they didn't require tqdm directly).

In order to fix it we will make a major version release, b2-sdk-python 2.0.0 - but before we do, we'll review the other public classes to make sure this type of a problem is not going to happen again with some other class.

mjurbanski-reef:
Continued in https://github.com/Backblaze/b2-sdk-python/pull/487

PR #398, opened 2023-06-14
fix: handle error with upload retry on unbound source

### LINKED ISSUE #394: `UnboundSourceBytes` can break on upload retry
When streaming data from unbound source (e.g. from `stdin`), it is possible (on a rare occasion) to perform a retry on a `IOWrapper` that was already released.

Currently the logic is as follows:
- a chunk of data is read from the stream;
- this chunk is wrapped into all the abstractions needed up to `WriteIntent`;
- it is ensured that at most N chunks are present at once;
- once `read` on a chunk goes to the end of the buffer, it's marked for removal – memory is reclaimed and new chunk can be downloaded.

This last point is problematic. If, during upload, the error happens after the very last write operation (the one that `read` last bytes from the write intent), a chunk is removed, and yet a retry is issued. Whenever a retry happens on an earlier `read` operation, there is no problem as the chunk is not marked for reclamation – only the very last `read` call is problematic.

The main problem is ensuring that no more than N buffers of data reside in memory at any given time while also ensuring that whenever a retry is issued, we have data to perform it with.

yedpodtrzitko:
- fixes #394

### DIFF
--- a/b2sdk/utils/typing.py
+++ b/b2sdk/utils/typing.py
@@ -12,4 +12,4 @@ from typing import TypeVar
 
 from b2sdk.transfer.outbound.upload_source import AbstractUploadSource
 
-TypeUploadSource = TypeVar("TypeUploadSource", bound=AbstractUploadSource)
+_TypeUploadSource = TypeVar("_TypeUploadSource", bound=AbstractUploadSource)
--- a/b2sdk/utils/typing.py
+++ b/b2sdk/utils/typing.py
@@ -1,0 +1,15 @@
+######################################################################
+#
+# File: b2sdk/utils/typing.py
+#
+# Copyright 2023 Backblaze Inc. All Rights Reserved.
+#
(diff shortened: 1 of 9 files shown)


b2sdk/utils/typing.py
```diff
@@ -0,0 +1,15 @@
...
+#
+# License https://www.backblaze.com/using_b2_code.html
+#
+######################################################################
+
+from typing import TypeVar
+
+from b2sdk.transfer.outbound.upload_source import AbstractUploadSource
+
+TypeUploadSource = TypeVar("TypeUploadSource", bound=AbstractUploadSource)
```
ppolewicz, line 15: should this become a part of the public interface of the library?

yedpodtrzitko: most likely not, gonna prefix it with underscore

yedpodtrzitko: done

PR #404, opened 2023-07-06
Typed lifecycle rules

mjurbanski-reef:


b2sdk/utils/docs.py
```diff
@@ -0,0 +1,39 @@
...
+#
+######################################################################
+from __future__ import annotations
+
+
+def get_b2sdk_doc_urls(type_: type) -> dict[str, str]:
+    """
+    Get the external documentation URLs for a b2sdk class.
+
+    Bon-b2sdk classes are not, and will not be supported.
```
mpnowacki-reef, line 17: what is `Bon-b2sdk classes` ?

mpnowacki-reef, line 17:
```suggestion
    Non-b2sdk classes are not, and will not be supported.
```

mpnowacki-reef (approved):

PR #412, opened 2023-08-09
File info by name

przadka:
I included the file info by name method to the API. This is related to https://github.com/Backblaze/B2_Command_Line_Tool/issues/440


mjurbanski-reef:
lets make sure this will be a Squash merge since a lot of code was reverted in these commits and it will make for noisy git history if  commits don't get squashed

Issue #357, opened 2022-09-23
How to set cache-control header when using upload_local_file


djbloc:
How do you set the X-Bz-Info-b2-cache-control HTTP header documented [here](https://www.backblaze.com/b2/docs/b2_upload_file.html) using the upload_local_file function?

I've tried to set via file_infos attribute using `file_infos={"Cache-Control": "max-age=86400"}` with no success. 

On testing using HEAD request to S3 URL returns

> HTTP/1.1 200 
> Accept-Ranges: bytes
> Last-Modified: Fri, 23 Sep 2022 18:38:23 GMT
> ETag: "079880d5addca95fa9846fa5033d72e1"
> x-amz-meta-cache-control: max-age=86400
> x-amz-request-id: 2a415cc03ebeade2
> x-amz-id-2: aMSlmgTPFZkBlojSzNYUznDk/Ym0w2TXK
> x-amz-version-id: _redacted_
> Content-Type: video/mp2t
> Content-Length: 268088
> Date: Fri, 23 Sep 2022 18:40:19 GMT

On testing using HEAD request to Friendly URL returns

> 
> HTTP/1.1 200 
> x-bz-file-name: _redacted_
> x-bz-file-id: _redacted_
> x-bz-content-sha1: 4251cdc9aaa9e22c76d5191004431b84e10ed773
> X-Bz-Upload-Timestamp: 1663958303158
> Accept-Ranges: bytes
> x-bz-info-cache-control: max-age=86400
> Content-Type: video/mp2t
> Content-Length: 268088
> Date: Fri, 23 Sep 2022 18:41:08 GMT

Note that neither approach returns a native "Cache-Control" HTTP header

ppolewicz:
As [documentation of Bucket.upload_local_file](https://b2-sdk-python.readthedocs.io/en/master/api/bucket.html#b2sdk.v2.Bucket.upload_local_file) shows, this is not implemented in b2-sdk-python yet. It seems that implementing it would just involve passing a parameter in a couple of places + two tests that would make sure this feature keeps working and is not broken by some future change. Would you like to take a stab at contributing a pull request to add it? b2-sdk-python developers provide [quite a lot of support](https://github.com/Backblaze/b2-sdk-python/blob/master/CONTRIBUTING.md) for contributors and you can learn quite a bit when working on it.

israel-oye:
[...]

PR #441, opened 2023-11-17
Header args v3

mpnowacki-reef:
This PR adds expire, content_disposition, content_encoding, content_language, arguments to various Bucket methods. It also fixes a bug related to cache_control handling in raw_api which caused cache_control parameter to not be propagated correctly on some operations.


mpnowacki-reef:
resolves https://github.com/Backblaze/b2-sdk-python/issues/357

PR #447, opened 2023-11-22
use xdg compliant path for config

jsuchan-reef:
- addressing Backblaze/B2_Command_Line_Tool/issues/958

- to avoid making this a breaking change and ignoring existing  config file in non-XDG path, there's a new step of checking for given file before checking XDG path. That makes the logic more consistent for both scenarios when `profile` is provided and when it's not.

- the same change in docstring needs to be done in B2 CLI repo, but let's validate the approach here first before duplicating the work

- in v4 we can eventually make a breaking change and remove the initial check of existing `B2_ACCOUNT_INFO_PROFILE_FILE` on XDG-like OSes before checking the XDG path.


test/unit/account_info/test_account_info.py
```diff
@@ -414,30 +415,24 @@ def _make_sqlite_account_info(self, env=None, last_upgrade_to_run=None):
...
 
-    def test_uses_default(self):
-        account_info = self._make_sqlite_account_info(
-            env={
-                'HOME': self.test_home,
-                'USERPROFILE': self.test_home,
-            }
-        )
-        actual_path = os.path.abspath(account_info.filename)
-        assert os.path.join(self.test_home, '.b2_account_info') == actual_path
```
jsuchan-reef, line 425: adjusting this test to make it pass made it moreless identical with `test_uses_xdg_config_home`, so it's removed

b2sdk/account_info/sqlite_account_info.py
```diff
@@ -91,10 +93,25 @@ def __init__(self, file_name=None, last_upgrade_to_run=None, profile: str | None
         )
 
     @classmethod
-    def _get_user_account_info_path(cls, file_name=None, profile=None):
+    def get_xdg_config_path(cls) -> str | None:
```
mjurbanski-reef, line 96: do we want to expose this method?
maybe it will be safer to hide it with `_`?

b2sdk/account_info/sqlite_account_info.py
```diff
@@ -91,10 +93,25 @@ def __init__(self, file_name=None, last_upgrade_to_run=None, profile: str | None
...
+        """
+        Return XDG config path if the OS is XDG-compatible (Linux, BSD), None otherwise.
+
+        If $XDG_CONFIG_HOME is empty but the OS is XDG compliant, fallback to ~/.config as expected by XDG standard.
+        """
+        xdg_config_home = os.getenv(XDG_CONFIG_HOME_ENV_VAR)
+        if xdg_config_home or sys.platform.startswith(('linux', 'freebsd', 'openbsd', 'netbsd')):
```
mjurbanski-reef, line 103: ideally we would use `platformdirs` package, but since I guess it doesn't quite make sense to add another 3rd party dependency just for hits, maybe lets take a lesson from them and make this into
```suggestion
        if xdg_config_home or not sys.platform == "win32":
```
https://github.com/platformdirs/platformdirs/blob/main/src/platformdirs/__init__.py#L24

Only suggesting that, since it seems it will be very easy to miss something, and `platformdirs` ppl are the one most likely to have get it right since a lot of packages rely on them https://github.com/platformdirs/platformdirs/network/dependents?dependent_type=PACKAGE

test/unit/account_info/test_account_info.py
```diff
@@ -414,30 +415,24 @@ def _make_sqlite_account_info(self, env=None, last_upgrade_to_run=None):
...
     def test_uses_xdg_config_home(self, apiver):
+        is_xdg_os = bool(SqliteAccountInfo.get_xdg_config_path())
         with WindowsSafeTempDir() as d:
             account_info = self._make_sqlite_account_info(
                 env={
                     'HOME': self.test_home,
                     'USERPROFILE': self.test_home,
-                    XDG_CONFIG_HOME_ENV_VAR: d,
+                    # pass the env. variable on XDG-like OS only
+                    XDG_CONFIG_HOME_ENV_VAR: is_xdg_os and d,
```
mjurbanski-reef, line 426: this will set `XDG_CONFIG_HOME_ENV_VAR` to `False`, while it doesn't change the outcome of the code, 
this test already relies to much on internal behavior, so lets fix at least this and only add `XDG_CONFIG_HOME_ENV_VAR` when needed to env

test/unit/account_info/test_account_info.py
```diff
@@ -414,30 +415,24 @@ def _make_sqlite_account_info(self, env=None, last_upgrade_to_run=None):
...
-            env={
-                'HOME': self.test_home,
-                'USERPROFILE': self.test_home,
-            }
-        )
-        actual_path = os.path.abspath(account_info.filename)
-        assert os.path.join(self.test_home, '.b2_account_info') == actual_path
-
     def test_uses_xdg_config_home(self, apiver):
+        is_xdg_os = bool(SqliteAccountInfo.get_xdg_config_path())
```
mjurbanski-reef, line 419: this worries me as, as we don't verify `get_xdg_config_path()`  properly recognizing XDG compatibility at any point

maybe instead of this please use your variation of platform xdg recognition here?
OR the one from `platformdirs` as test-only dependency?

mjurbanski-reef (approved):

test/unit/account_info/test_account_info.py
jsuchan-reef: adding another test for that

b2sdk/account_info/sqlite_account_info.py
jsuchan-reef: their decision logic seems to be `sys.platform not in ('win32', 'linux')`, we can use that without adding the whole package as a dependency, that feels like an overkill.

PR #456, opened 2023-12-12
Change auhtorize account signature

mpnowacki-reef:


z1lc:
Isn't this a breaking (backwards-incompatible) change? I would have expected this to increment the major version since users that call this method using positional arguments will have their integration break.

ppolewicz:
It is a breaking change for apiver v3 (which is not to be used yet, it is considered to be an unstable interface). For v2 a wrapper has been added that restores the compatibility. If you properly import from b2sdk.v2, your code will work as it did before. One day when you'll be switching to b2sdk.v3 (when _you_ decide you are ready), you'll either have to copy the wrapper from b2sdk.v2 to your code or adjust your code to use b2sdk.v3.

After b2sdk.v3 is released, we'll expect all new users (as well as maybe old users writing new programs) to use b2sdk.v3 and to use the new interface, where `realm` is no longer a mandatory argument. In practice this would only be used if you had access to a non-production instance of B2 server, which probably means you are working for Backblaze. Since a small portion of the userbase needs to use this argument, it makes sense to provide a default value which almost everyone will use, rather than forcing everyone to set it explicitly.

Because we use these compatibility wrappers, we are able to modernise our function signatures, rename classes, deprecate old ways in favor of completely redesigned interfaces - all without breaking compatibility with old software. It's been so many years and we are still at major version 1, changed a ton of stuff (you can observe it all in b2sdk.v1, b2sdk.v2, b2sdk._v3) without breaking compatibility or abandoning users of old versions (except deprecating EOL versions of python - but nowadays in most cases you can migrate to new python using docker even if your operating system stays old).

Issue #458, opened 2024-01-14
Invalid escape sequences in docstrings


ehossack:
When running the library with python 3.12, I see the following:

```
/usr/local/lib/python3.12/site-packages/b2sdk/file_version.py:219: SyntaxWarning: invalid escape sequence '\_'
"""
/usr/local/lib/python3.12/site-packages/b2sdk/raw_simulator.py:1162: SyntaxWarning: invalid escape sequence '\?'
'/b2api/v[0-9]+/b2_download_file_by_id\?fileId=(?P<file_id>[^/]+)',
usr/local/lib/python3.12/site-packages/b2sdk/v2/file_version.py:27: SyntaxWarning: invalid escape sequence '\_'
```

PR #459, opened 2024-01-14
fix invalid escape sequences (python 3.12 SyntaxWarning)

ehossack:
I don't see any reason to not do this (nothing in sphinx?).

As referenced in https://github.com/Backblaze/b2-sdk-python/issues/458

### DIFF
--- a/b2sdk/raw_simulator.py
+++ b/b2sdk/raw_simulator.py
@@ -1159,7 +1159,7 @@ class RawSimulator(AbstractRawApi):
     DOWNLOAD_URL_MATCHER = re.compile(
         DOWNLOAD_URL + '(?:' + '|'.join(
             (
-                '/b2api/v[0-9]+/b2_download_file_by_id\?fileId=(?P<file_id>[^/]+)',
+                r'/b2api/v[0-9]+/b2_download_file_by_id\?fileId=(?P<file_id>[^/]+)',
                 '/file/(?P<bucket_name>[^/]+)/(?P<file_name>.+)',
             )
         ) + ')$'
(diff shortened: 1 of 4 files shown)


ehossack:
(I could add a changelog section, but not sure what version this would target. 1.30.0? 1.29.1?

b2sdk/raw_simulator.py
```diff
@@ -1159,7 +1159,7 @@ class RawSimulator(AbstractRawApi):
     DOWNLOAD_URL_MATCHER = re.compile(
         DOWNLOAD_URL + '(?:' + '|'.join(
             (
-                '/b2api/v[0-9]+/b2_download_file_by_id\?fileId=(?P<file_id>[^/]+)',
+                '/b2api/v[0-9]+/b2_download_file_by_id?fileId=(?P<file_id>[^/]+)',
```
mjurbanski-reef, line 1162: wow, this seems like this regexp was invalid from the beginning, right?

instead "fix" should be just adding `r` i.e. `r'...'` as otherwise this `?` will mean `d` is optional

mjurbanski-reef:
As for the changelog, we use towncrier to manage main file, so you just create a new one and don't worry about versions yourself.
Please instructions in  https://github.com/Backblaze/b2-sdk-python/blob/master/CONTRIBUTING.md#changelog .

P.S. thanks for contributing

b2sdk/raw_simulator.py
ehossack: good point.

ehossack:
> @ehossack the changelog files need to follow a specific naming scheme. ~If it's a fix, it should be `+<unique-change-name>.fixed.md`, e.g. `+syntax_warning.fixed.md`. Once you rename the file, it should be good to go. :)~
> 
> _Edit_: since this fixes issue #458, the name of the changelog file should be `458.fixed.md`.

Sorry, new process and also I didn't properly read the section you linked 😅

mjurbanski-reef (approved):
looks great, thank you!

PR #469, opened 2024-02-01
Fix invalid escape sequences

vbaltrusaitis-reef:


mjurbanski-reef:
fixed in other PR already

PR #475, opened 2024-02-11
Add type annotations to api methods for improved type checking

przadka:


b2sdk/api.py
```diff
@@ -386,7 +386,8 @@ def delete_bucket(self, bucket):
         account_id = self.account_info.get_account_id()
         self.session.delete_bucket(account_id, bucket.id_)
 
-    def list_buckets(self, bucket_name=None, bucket_id=None, *, use_cache: bool = False):
+    def list_buckets(self, bucket_name=None, bucket_id=None, *,
+                     use_cache: bool = False) -> list[Bucket]:
```
ppolewicz, line 390: Shouldn't this be a `collections.abc.Sequence` rather than a list?

PR #479, opened 2024-03-05
Remove setuptools as a runtime dependency

vishwin:
`setuptools` is not actually used in any runtime code. Only `liccheck` during `lint` uses it for `pkg_resources`, which has long been deprecated (and is more `liccheck`'s problem to migrate away). `liccheck` also does not yet officially support Python 3.12, otherwise I would think they would explicitly specify the `setuptools` dependency themselves.

The only reason why `setuptools` would be present in a runtime environment is for `pkg_resources`, whose functionality has been replaced by `packaging` and the `importlib` suite. Otherwise, `setuptools` pollutes the environment, particularly at the operating system distribution level. Many Python packages have not adopted PEP 517 and still execute `setup.py` directly, functionality that `setuptools` deprecated in 58 and no longer supports after. Multiple `setuptools` cannot exist in the same environment.

### DIFF
--- a/changelog.d/+setuptools.removed.md
+++ b/changelog.d/+setuptools.removed.md
@@ -1,0 +1,1 @@
+Remove `setuptools` as a runtime dependency; retained for `lint` as `liccheck` still uses `pkg_resources`.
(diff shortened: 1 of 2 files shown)


mjurbanski-reef (approved):

mjurbanski-reef:
seems like leftover from old days, thank you for fixing this!

PR #512, opened 2024-10-25
upgrade to pytest 8

selfisekai:


ppolewicz:
@selfisekai yapf would like to mess up your pretty formatting for the sake of consistency. Could you please add a `# yapf: disable` at the end of the affected lines to make sure your formatting is not vandalized by a bot?

Also, it appears that you've changed quotes in a bunch of lines. Do we really need to do it in this PR?

selfisekai:
i've run ruff over it, that's how it happened

mjurbanski-reef (approved):
the `pytest-lazy-fixtures` library is new and unknown, so I went and checked github repo and current tar/wheel releases. They seem clean - i.e. nothing shady going in there, but before updating the lockfile similar review should be done to make sure it's not a security hazard to dev machines.

PR #514, opened 2024-10-25
pdm: solve the package-type warning

selfisekai:


mjurbanski-reef:
Changelog message is missing, but in this particular case I think it should be okay.

PR #520, opened 2024-10-30
pytest: switch hook from path to collection_path

selfisekai:
solves PytestRemovedIn9Warning: The (path: py.path.local) argument is deprecated, please use (collection_path: pathlib.Path) https://docs.pytest.org/en/latest/deprecations.html#py-path-local-arguments-for-hooks-replaced-with-pathlib-path


mlech-reef:
lgtm

Issue #439, opened 2023-11-16
b2sdk UploadTokenUsedConcurrently


mjurbanski-reef:
To utilize b2sdk under asyncio, I'm running it under threadpool executor, but strangely enough, error occurred once per server and not multiple times on a single server (I have ~80 of them running this code, and they upload a file every other minute). Error occurred on around `2023-11-16T04:59:14.000Z`  on 3 servers. Posting that info, since all my servers reported the same token in the exception, which seems to indicate b2 API server bug.

```
UploadTokenUsedConcurrently: More than one concurrent upload using auth token 4_005fREDACTED_01b061e5_0b1c7c_uplg_-OREDACTEDgux4J-REDACTED4hORgY=
  File "concurrent/futures/thread.py", line 58, in run
    result = self.fn(*self.args, **self.kwargs)
  File "myapp/worker/tasks_b2.py", line 179, in process_datasource_dir
    return _process_datasource_dir(datasource_dir, claimed_datasource_ids, datasource_id=datasource_id)
  File "myapp/worker/tasks_b2.py", line 219, in _process_datasource_dir
    chunk_data = _upload_merged_chunks(
  File "myapp/worker/tasks_b2.py", line 242, in _upload_merged_chunks
    upload_url = process_upload(BUCKET_dataS, data_file_merged, datasource_id, data_bucket_path)
  File "myapp/worker/tasks_b2.py", line 96, in process_upload
    upload_result = bucket.upload_local_file(
  File "logfury/_logfury/trace_call.py", line 86, in wrapper
    return function(*wrapee_args, **wrapee_kwargs)
  File "b2sdk/version_utils.py", line 116, in wrapper
    return func(*args, **kwargs)
  File "b2sdk/v2/bucket.py", line 83, in upload_local_file
    return super().upload_local_file(
  File "logfury/_logfury/trace_call.py", line 86, in wrapper
    return function(*wrapee_args, **wrapee_kwargs)
  File "b2sdk/bucket.py", line 592, in upload_local_file
    return self.concatenate(
  File "logfury/_logfury/trace_call.py", line 86, in wrapper
    return function(*wrapee_args, **wrapee_kwargs)
  File "b2sdk/bucket.py", line 1018, in concatenate
    return self.create_file(
  File "logfury/_logfury/trace_call.py", line 86, in wrapper
    return function(*wrapee_args, **wrapee_kwargs)
  File "b2sdk/bucket.py", line 842, in create_file
    return self._create_file(
[...]

PR #522, opened 2024-10-31
Add upload token reset after upload timeout

ppolewicz:
Probably fixes https://github.com/Backblaze/b2-sdk-python/issues/439


ppolewicz:
You know this feeling when you write code and then people still find bugs in it 6 years later?

The fix in a race-condition-infested environment was found to be effective by the reporter.

PR #535, opened 2025-03-26
EmergePlanner: Optimize _get_emerge_parts() for better grouping of parts in the upload buffer

LewyMaster:
EmergePlanner optimization solution as a task for job recruiting process.

Task:
`implementation of https://github.com/Backblaze/b2-sdk-python/blob/10a5e720237cf2581a7f6297eda13f0e725a685b/b2sdk/transfer/emerge/planner/planner.py#L191
is a little bit naiive, because (with 5MB minimum part size) if we mark
"u" as a megabyte of data to upload
"c" as a megabyte of data to copy
"d" as a megabyte of data that will be downloaded and uploaded
then for "input", "master" pattern will be executed by the current version of the code while ideally in such case the "ideal" pattern would be used:
# input:  uuu  cccccccccc  uu  ccccc ...
# master: uuu  ddcccccccc  uu  ddddd ...
# ideal:  uuu  ddcccccddd  uu  ccccc ...
This is because Planner is only considering to bundle an upload with a part of a copy from the right side of the upload and not also from the left. Appropriately to the quality of the code in this project, tests should be provided and the code of the planner should remain readable. The behavior of the planner should strictly improve and the changes done to it should not have any negative impact on any input case (beyond marginally higher CPU consumption during planning).
`

### DIFF
--- a/b2sdk/transfer/emerge/planner/planner.py
+++ b/b2sdk/transfer/emerge/planner/planner.py
@@ -176,26 +176,51 @@ class EmergePlanner(object):
                     missing_length = min_part_size - upload_buffer.length
                 else:
                     missing_length = 0
-                if missing_length > 0 and current_len - missing_length < min_part_size:
-                    # current intent is *not* a "small copy", but upload buffer is small
-                    # and current intent is too short with the buffer to reach the minimum part size
-                    # so we append current intent to upload buffer
-                    upload_buffer.append(current_intent, current_end)
-                else:
-                    if missing_length > 0:
-                        # we "borrow" a fragment of current intent to upload buffer
-                        # to fill it to minimum part size
-                        upload_buffer.append(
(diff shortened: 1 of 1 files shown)


mzukowski-reef:
we were discussing back then splitting implementations for streaming and "full knowledge" planner - then we would be able to plan better in such edge cases - but this is only for small parts - so we decided to keep single implementation and leave such edge cases suboptimal - in most applications parts will be larger than 5MB and there is no problem - if there is valid usecase thenI think dedicated implementation of planner should be produced for "full knowledge" case - modifying streaming logic is nightmare :)

ppolewicz (commented):
I didn't read it in detail, because it doesn't have a single test yet. The emerger planner has very nice tests, if I remember correctly - could you please add one showing the case described in task description, make sure that it does what it should do?

ppolewicz:
It was abandoned

PR #548, opened 2025-10-12
Faster upload token refresh on 5xx errors during uploads

ppolewicz:
When a server returns a 5xx error (e.g., 503 Service Unavailable) during file upload, the SDK now properly refreshes upload tokens instead of retrying with stale cached upload URLs a bunch of times first.

Previously, the HTTP layer would catch ServiceError and retry using the same pre-fetched upload URLs from the token pool. This caused repeated failures because the upload URLs may no longer be valid after a 5xx error and it delayed the recovery.

The fix introduces ServiceErrorDuringUpload exception that disables HTTP-level retries (similar to B2RequestTimeoutDuringUpload), forcing the error to propagate to the upload manager. The upload manager then clears cached tokens via clear_bucket_upload_data() and requests fresh upload URLs for retry attempts.

Changes:
- Added ServiceErrorDuringUpload exception class
- Convert ServiceError to ServiceErrorDuringUpload in b2http.py
- Fixed RawSimulator to generate unique upload tokens per URL
- Added comprehensive test for token refresh on 503 errors

Fixes Backblaze/B2_Command_Line_Tool#1118

### DIFF
--- a/test/unit/test_upload_503_token_refresh.py
+++ b/test/unit/test_upload_503_token_refresh.py
@@ -18,11 +18,8 @@ from __future__ import annotations
 
 from unittest.mock import patch
 
-import pytest
-
 from b2sdk._internal.exception import ServiceError
-from b2sdk.v3 import RawSimulator
-from b2sdk.v3 import B2Api, B2HttpApiConfig, DummyCache, StubAccountInfo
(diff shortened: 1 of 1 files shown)


olzhasar-reef:
Closing in favor of #556

PR #574, opened 2026-05-21
Retry integration tests on transient service errors

ppolewicz:
This PR verifies integration retry behavior against live B2 CI credentials.\n\nIt adds integration-test retries for transient B2 service errors and an upload test that uses the documented X-Bz-Test-Mode: fail_some_uploads hook.

### DIFF
--- a/test/integration/conftest.py
+++ b/test/integration/conftest.py
@@ -14,7 +14,7 @@ import pytest
 from b2sdk._internal.exception import ServiceError
 
 RETRYABLE_SERVICE_ERROR_STATUSES = {500, 503}
-INTEGRATION_TEST_RETRY_COUNT = 2
+INTEGRATION_TEST_RETRY_COUNT = 4
 
 
 @pytest.fixture(scope='session', autouse=True)
--- a/test/integration/conftest.py
+++ b/test/integration/conftest.py
@@ -11,7 +11,7 @@ from __future__ import annotations
 
 import pytest
 
-from b2sdk._internal.exception import ServiceError
+from b2sdk._internal.exception import ServiceError, TooManyRequests
 
 RETRYABLE_SERVICE_ERROR_STATUSES = {500, 503}
 INTEGRATION_TEST_RETRY_COUNT = 4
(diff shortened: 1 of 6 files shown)


test/integration/conftest.py
```diff
@@ -7,9 +7,44 @@
...
+
+@pytest.hookimpl(tryfirst=True)
+def pytest_pyfunc_call(pyfuncitem):
+    testfunction = pyfuncitem.obj
+    funcargs = pyfuncitem.funcargs
+    testargs = {arg: funcargs[arg] for arg in pyfuncitem._fixtureinfo.argnames}
+
+    for attempt in range(INTEGRATION_TEST_RETRY_COUNT + 1):
+        try:
+            testfunction(**testargs)
```
jan-sauer-reef, line 33: There are no async tests but if there ever were in the future, they would never be awaited and unless the test runner sees the runtime warning, they'd never know -> they'd appear to always pass. It might be worth checking that `testfunction` isn't async.

test/integration/conftest.py
```diff
@@ -7,9 +7,44 @@
 # License https://www.backblaze.com/using_b2_code.html
 #
 ######################################################################
+from __future__ import annotations
+
 import pytest
 
+from b2sdk._internal.exception import ServiceError, TooManyRequests
+
+RETRYABLE_SERVICE_ERROR_STATUSES = {500, 503}
```
jan-sauer-reef, line 16: Tests are retried on 500 errors but the bucket operations in `_retry_bucket_test_operation()` only on 503.

a) Is nested retry logic necessary for a test?
b) If so, shouldn't the BucketManager also retry on 500 errors then?

jan-sauer-reef (changes requested):
Nice! Unless async integration tests are categorically out of the question, though, I would add a provision in the `pytest_pyfunc_call` function to await async tests, otherwise they'll pass almost-silently (albeit with a RuntimeWarning).

test/integration/conftest.py
ppolewicz: there is not a shred of `async` in this project

test/integration/conftest.py
ppolewicz: you are right, it should always retry on 5xx

PR #576, opened 2026-05-25
Expose `b2_get_download_authorization` response header overrides

### LINKED ISSUE #539: b2_get_download_authorization extra parameters
Hi,

`/b2api/v4/b2_get_download_authorization` accepts many parameters that are not exposed in this lib. All b2* are missing.

I am specifically looking for `b2ContentDisposition` when generating auth token.

Can I make a PR that adds them to `get_download_authorization` method, or is there a reason its not included?

Thank you

Montana:
Add keyword-only parameters for the optional `b2ContentDisposition`, `b2ContentLanguage`, `b2Expires`, `b2CacheControl`, `b2ContentEncoding`, and `b2ContentType` fields accepted by `b2_get_download_authorization`.

The fields are omitted from the request unless explicitly provided, preserving existing behavior.

Fixes #539.

### DIFF
--- a/b2sdk/_internal/raw_api.py
+++ b/b2sdk/_internal/raw_api.py
@@ -288,7 +288,19 @@ class AbstractRawApi(metaclass=ABCMeta):
 
     @abstractmethod
     def get_download_authorization(
-        self, api_url, account_auth_token, bucket_id, file_name_prefix, valid_duration_in_seconds
+        self,
+        api_url,
+        account_auth_token,
+        bucket_id,
+        file_name_prefix,
+        valid_duration_in_seconds,
+        *,
+        b2_content_disposition: str | None = None,
+        b2_content_language: str | None = None,
+        b2_expires: str | None = None,
+        b2_cache_control: str | None = None,
+        b2_content_encoding: str | None = None,
+        b2_content_type: str | None = None,
     ):
         pass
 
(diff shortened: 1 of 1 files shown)


ppolewicz:
Thanks for putting this together @Montana - exposing the response-header overrides on `get_download_authorization` is a genuinely useful addition and the request-construction logic looks correct. A few things need to be addressed before this can be merged:

**1. The simulator wasn't updated.** This change touches the abstract `AbstractRawApi.get_download_authorization` signature and the `B2RawHTTPApi` implementation, but `RawSimulator` implements the same interface and still has the old signature. As it stands, any code (including the test suite) that exercises this method against the simulator with the new parameters will raise a `TypeError`. The simulator needs to accept and honor the same keyword-only params so the emulated behavior stays in sync with the real API. This way users of the sdk will be able to test their code without touching the cloud and without swimming in a sea of mocks.

**2. No tests.** We need coverage for the new behavior: that each field is forwarded when provided, that it's omitted from the request when left as `None` (the existing-behavior guarantee you mention in the description), and a round-trip through the simulator.

**3. No changelog entry or documentation.** Please add a changelog entry and document the new parameters where the method is described, so users can actually discover them. I think this should include the description of the behavior change, so that the users of the sdk don't have to go through the documentation of the API and through RFCs which define the expected client behavior.

**4. Parameter naming.** The `b2_content_disposition` / `b2_content_language` / `b2_expires` / etc. names are direct snake_case transliterations of the raw JSON keys, `b2_` prefix and all. The `b2ContentDisposition` form belongs in the wire payload (where you've correctly used it), but the Python-facing parameters should follow the SDK's conventions and drop the redundant prefix — `content_disposition`, `content_language`, `expires`, `cache_control`, `content_encoding`, `content_type`. The fact that they're B2 parameters is already established by the method and the package.

One broader note: `raw_api` is the lowest layer, so even once the above is fixed, callers won't see these options until they're threaded through the public-facing methods (e.g. `Bucket.get_download_authorization`).

Happy to review again once these are in.

Montana:
Hey @ppolewicz,

Made PR's for `bucket.py` and `session.py` will be finishing up the rest later.

Cheers,
Michael Mendy

ppolewicz:
It should all be in one PR that adds those headers as well as relevant tests

PR #577, opened 2026-05-26
Add Read the Docs build config

ppolewicz:
## Summary
- add a checked-in Read the Docs config for this repository
- install the system packages required by the Sphinx docs build on RTD
- install the Python doc dependency group with `uv` so RTD builds match local/CI docs builds

## Why
The repo currently builds docs in CI, but CD does not publish docs and the repository does not include a `.readthedocs.yaml` file for RTD to consume. The docs build also requires `graphviz` and `plantuml`, which are installed in GitHub Actions but were not declared for RTD.

## Verification
- `uv run nox --non-interactive -vs doc`

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -1,0 +1,20 @@
+version: 2
+
+build:
+  os: ubuntu-24.04
+  tools:
+    python: "3.14"
+  apt_packages:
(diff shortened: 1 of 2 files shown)


jan-sauer-reef (approved):
👍

PR #578, opened 2026-06-01
Enhance get_download_authorization with optional parameters

Montana:
`get_download_authorization` method to include additional optional parameters for better customization.


ppolewicz:
duplicate of #576

PR #579, opened 2026-06-01
feat(bucket): expose download authorization response-header overrides

Montana:
Add optional response-header override parameters to `Bucket.get_download_authorization()` and forward them through the session layer.

Supported overrides include content disposition, language, expiration, cache control, encoding, and content type.


ppolewicz:
duplicate of #576

PR #581, opened 2026-07-02
Retry stream read errors in SimpleDownloader

agoncharov-reef:
`SimpleDownloader` already had range-based retry logic for truncated downloads (e.g. when the server closes the connection mid-stream), but `ChunkedEncodingError`, `ConnectionError`, and `ContentDecodingError` raised during `iter_content` bypassed that path and failed the download immediately.

This PR catches those exceptions during both the initial read and follow-up range requests, logs them at debug level, and lets the existing retry loop resume. Retry is skipped when content is decoded (`content_encoding` set and `decode_content=True`), since a partially decoded stream cannot be resumed with byte ranges.

Also fixes `v1.B2Api` not exposing `api_config`, and ensures decoded downloads still get truncation checks after download (hash verification remains skipped for them).

### DIFF
--- a/b2sdk/_internal/file_version.py
+++ b/b2sdk/_internal/file_version.py
@@ -471,6 +471,10 @@ class DownloadVersion(BaseFileVersion):
             return None
         return parse_http_date(self.expires)
 
+    @property
+    def _should_be_decoded(self) -> bool:
+        return bool(self.content_encoding and self.api.api_config.decode_content)
+
     def as_dict(self) -> dict:
         result = super().as_dict()
         if self.cache_control is not None:
--- a/b2sdk/_internal/transfer/inbound/downloader/abstract.py
+++ b/b2sdk/_internal/transfer/inbound/downloader/abstract.py
@@ -118,8 +118,7 @@ class AbstractDownloader(metaclass=B2TraceMetaAbstract):
             return False
         if (
             not self.SUPPORTS_DECODE_CONTENT
-            and download_version.content_encoding
-            and download_version.api.api_config.decode_content
+            and download_version._should_be_decoded
         ):
             return False
         return True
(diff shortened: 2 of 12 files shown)


b2sdk/_internal/transfer/inbound/downloader/abstract.py
```diff
@@ -116,11 +116,7 @@ def is_suitable(self, download_version: DownloadVersion, allow_seeking: bool):
         """
         if self.REQUIRES_SEEKING and not allow_seeking:
             return False
-        if (
-            not self.SUPPORTS_DECODE_CONTENT
-            and download_version.content_encoding
-            and download_version.api.api_config.decode_content
-        ):
+        if not self.SUPPORTS_DECODE_CONTENT and download_version._should_be_decoded:
```
ppolewicz, line 119: `_should_be_decoded` is private and should not be accessed from outside. Rename to `should_be_decoded`?

b2sdk/_internal/transfer/inbound/downloader/simple.py
```diff
@@ -58,33 +64,35 @@ def _download(
         # or something and the server closes connection, while neither tcp or http have a problem
         # with the truncated output, so we detect it here and try to continue
 
-        num_tries = 5  # this is hardcoded because we are going to replace the entire retry interface soon, so we'll avoid deprecation here and keep it private
```
ppolewicz, line 61: I don't think you should un-hardcode it here as when this will turn into an interface that folks can rely on, it will be difficult to maintain the interface over the refactoring. Move to module scoped `_MAX_RETRIES`?

b2sdk/_internal/file_version.py
```diff
@@ -471,6 +471,10 @@ def expires_parsed(self) -> dt.datetime | None:
             return None
         return parse_http_date(self.expires)
 
+    @property
+    def _should_be_decoded(self) -> bool:
```
ppolewicz, line 475: if the rest of the code expects `DownloadVersion`-like objects to have this method or else it's going to crash, then we have to add it to interface, meaning we have to document it so that those who inherit will know what it is

